PR #8, opened 2016-07-17
swedish translation

Sopor-:


tibbi:
hello @Sopor- , how would you translate "Dark theme"? Thanks!

Sopor-:
@tibbi "Mörkt tema" and if you want "Light theme" it will be "Ljust tema"

PR #36, opened 2016-10-24
Translated some strings

mueller-ma:


tibbi:
@mueller-ma can you please translate a freshly added string, "Seems like your calendar is empty. You can create events with the Plus button at the bottom." ? Thanks!

PR #131, opened 2017-03-15
update Russian

mcbaze:
what about translatable changelog? have you decided to refuse this idea?


tibbi:
Ive made the release notes non translatable, as the translations are obviously always late anyway and the extra strings just scare away new translators. The "What's New" dialog is shown only once after upgrade and I usually type it just before releasing so theres no time for translating. And I dont think that whole CHANGELOG.md is worth translating.

PR #152, opened 2017-04-17
update german translation

nopeppermint:
Is there a way to see how these translation strings are used without building this app on my own?
I am not sure if my latest translation is correct in every use case ..


tibbi:
nope, especially these latest strings are not used anywhere yet.
After you click on "Repeat till" you will get another dialog where you can choose either "Repeat till a date" or "Repeat x times". After you select "Repeat x times" there will be another dialog that will contain just one textfield with a number with "times" after that.
I might make it simpler by having "Repeat till a date" selected initially with a button for "Repeat x times" somewhere, we will see.

Issue #163, opened 2017-05-02
missing days and repeated days in weekly view


zikobad:
some days are repeated and some days are missing in weekly view

1. repeated days
![screenshot_2017-05-02-15-03-37](https://cloud.githubusercontent.com/assets/20209910/25621253/c624e272-2f48-11e7-8274-09d616540e9e.png)
![screenshot_2017-05-02-15-03-41](https://cloud.githubusercontent.com/assets/20209910/25621254/c6267362-2f48-11e7-8630-c166a3bdf15a.png)

2. missing days

![screenshot_2017-05-02-15-03-50](https://cloud.githubusercontent.com/assets/20209910/25621252/c61e934a-2f48-11e7-865f-38eca40637de.png)
![screenshot_2017-05-02-15-03-53](https://cloud.githubusercontent.com/assets/20209910/25621250/c61b0e14-2f48-11e7-8ec9-ca0a17170173.png)

PR #164, opened 2017-05-02
handle time-zone change in weekly-view #163 

zikobad:
fix missing / repeated days in weekly views problem #163 .
problem : time zone change make the weektimestamp move to next/previous day .


tibbi:
nice one, thanks

PR #177, opened 2017-06-02
Update Russian translation

antonv6:
Bumped completeness to 90%+.

On a related note, constructing word-by-word phrases like "Every second Tuesday" will never look good in certain languages, like Russian, because days have different gender and "second" is a gender-dependent word (needs different forms for e.g. Tuesday and Wednesday).


tibbi:
Ye Im not sure how are other projects handling it, should be good enough. Ive added the extra strings because of the Slovak language.

PR #187, opened 2017-07-03
Use local country names for the list of holidays

antonv6:
The list already uses local names in local script for some countries, but not for all. Two names stood out particularly: Rossiya and Nihon. I can say that the former didn't look right at all, and the latter looked like it also needed improving, so I went ahead and changed both of them to be in local script.

https://en.wikipedia.org/wiki/Russia
https://en.wikipedia.org/wiki/Japan

Now, sometimes people want to add holidays of foreign countries, but don't understand their script, and this PR obviously makes this problem somewhat worse, but if that's important, then the whole list needs be made translatable as a full-fledged part of UI (or maybe Android already has translated country names somewhere).


tibbi:
okay thanks, I dont remember now where did I get the country names from :)
In the future I also want to have holiday names in both english, and in the native language.

PR #200, opened 2017-08-11
Polish translation

gregory678:
I have some doubts about some lines, but we'll see how it looks anyway.


tibbi:
hah nice, thanks. Just ask if you arent sure about some strings usage.

PR #202, opened 2017-08-12
Add LeakCanary for debug versions of the app

jonathan-caryl:
Android apps often leak memory, and Leak Canary let's us find those leaks. Impressively it found no leaks in Simple-Calendar, but it could be worthwhile adding for debug builds only to prevent accidental leaks being introduced in the future.


app/src/debug/kotlin/com/simplemobiletools/calendar/BuildVariantApplication.kt
```diff
@@ -0,0 +1,14 @@
+package com.simplemobiletools.calendar
+
+import com.squareup.leakcanary.LeakCanary
+
+open class BuildVariantApplication : BaseApp() {
```
jonathan-caryl: This is a slightly more complicated approach than just using the `leakcanary-android-no-op` version for release builds, but it's useful to be able to install some libraries only in debug builds, and not all of them come with appropriate no-op versions.

tibbi:
nice, thanks. Seems like you know what are you doing and theres really no reason why I couldnt include it in the debug build. 
We will just have to add some trick to exclude it on F-droid builds at https://f-droid.org . Its a platform where they build open source apps themselves from the given source, to guarantee that the .apk really contains the given code. As far as I know they dont sign their apps and they use debug builds, so people might get the version with leak canary included. We might have to create some flag for their builds or what, needs some research.

jonathan-caryl:
Interesting, I'll check it out.

tibbi:
you can take Amaze File Manager as an example, they have some exceptions for F-droid builds for example at the build.gradle, the project is at https://github.com/arpitkh96/AmazeFileManager . Their build script related things are at https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/com.amaze.filemanager.txt , calendar is at https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/com.simplemobiletools.calendar.txt

jonathan-caryl:
Having looked at the Amaze File Manager source a little bit, it seems like they have two product flavours, one that's compiled for Google Play, which uses Google Play Services, and one that's compiled for fdroid that doesn't (GPS being unavailable on a stock AOSP version of Android). It doesn't to me seem likely that you'd need to do the same to avoid using LeakCanary on an fdroid build, as LeakCanary is open source as well, but it could certainly be done.

tibbi:
My goal is not using LeakCanary on f-droid builds, to avoid those freezing popups.

PR #204, opened 2017-08-13
Add fdroid buildType so leak canary can be excluded.

jonathan-caryl:
I'm not sure if this is 100% complete, but I think it's along the lines of what you're looking for. Switch to having two productFlavors, one for fdroid, one regular. For the fdroid version the debug/release version of `BuildVariantAppliction` is the same and neither includes LeakCanary. For the regular one the debug version includes LeakCanary as before. Nothing else is different between the two productFlavors.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -25,12 +25,16 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
             signingConfig signingConfigs.release
         }
+        fdroid {
+            initWith debug
+        }
     }
 
     sourceSets {
         main.java.srcDirs += 'src/main/kotlin'
         debug.java.srcDirs += 'src/debug/kotlin'
         release.java.srcDirs += 'src/release/kotlin'
+        fdroid.java.srcDirs += 'src/fdroid/kotlin'
     }
 
     lintOptions {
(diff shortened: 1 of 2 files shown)


tibbi:
no need to prepend the existing flavors with `regular`, just add the new `fdroid` one. They use just a kind of debug build without proguard. So it will be 3 flavors: debug, release, fdroid. I will then update the fdroid build script to use the fdroid one.

jonathan-caryl:
Simplified as you suggested, just adds an `fdroid` buildType, which uses `initWith` to pick up its configuration from `debug` — so it should be easy to keep the two types in sync. And then the `BuildVariantApplication` for `fdroid` doesn't mention `LeakCanary` like the `release` version.

tibbi:
ye, that looks nice. I wasnt aware of the `initWith` thing, I dont think theres a cleaner way to do it. Ill merge it and try updating the fdroid build once Im done with caldav sync. Thanks

tibbi:
f-droid guys told me that they always build the release mode, its just an f-droid flavor that we could add, if needed. LeakCanary isnt active in the release mode tho, so its not needed. Ive removed the f-droid build variant related things, thanks anyway tho :)

jonathan-caryl:
no problem

PR #208, opened 2017-08-22
Add korean translation

yh-kim:


tibbi:
nice, thanks :) Some strings are missing, do you miss some context of where are they used?

tibbi:
alright, thanks

PR #250, opened 2017-10-23
Adding some missing XML declarations

RyDroid:


tibbi:
whats the use of it anyway?

RyDroid:
It is XML conformity. Some parser may not like this lack (crashing or wondering what is the correct charset and/or XML version) and some type auto-detect tools may not find the type without that.

PR #257, opened 2017-10-27
[l10n] Add breton translations

tornoz:
Add breton (br) locale translation file


tibbi:
Nice, thanks. However playstore doesnt list the Breton language, so I cannot use the playstore listing related strings for now :p

PR #290, opened 2017-12-06
Update norway.ics

FTno:
I have extended the file for norwegian public holidays with the years 2021-2024. I have also added 5 fixed public holidays at the bottom.
I do not understand how to make a value for the 'UID:' string, so I have just left the line empty after 'UID:'.
Could you add something there or if it is easy tell me how and I can make a new pull request.


tibbi:
cool, thanks. The UID is just a random string, there isnt any particular login in it. At least Im not aware of it. The goal of it is just to avoid duplicates, for example if you import the holidays twice, they will still be visible only once because I check if the given uid is free.

PR #292, opened 2017-12-06
Update norway.ics

FTno:
Added UID value and removed 'CATEGORIES:Holidays' compared to the previous pull request.


tibbi:
ye looks good, thanks

PR #302, opened 2018-01-01
Update Spanish translation

esu23:


tibbi:
Hey, thanks! I wrote a guide how to set up a file watcher which will send you an email whenever a file changes, you can check it out at https://github.com/SimpleMobileTools/General-Discussion#how-can-i-get-notified-if-a-translated-file-changes . That way you won't have to keep checking files over and over.

Issue #368, opened 2018-02-18
All-day CalDAV events show up one day early


angelsl:
Caused by 91d5c3dd00e16320ede02b71ca07eb7acd1f9fa7

I'm not sure what the intention was behind that commit (is there some calendar sync adapter that is inserting non-UTC timestamps into the calendar provider?), but this breaks Google account calendar all-day events _if the user's timezone is greater than GMT+5_:

Google has `startTS` set to midnight UTC on the day itself, and `endTS` set to midnight UTC on the day after.

([CalDAVHandler.kt L237](https://github.com/SimpleMobileTools/Simple-Calendar/blob/master/app/src/main/kotlin/com/simplemobiletools/calendar/helpers/CalDAVHandler.kt#L237))

```kt
// 1
if (allDay == 1 && timeZone == "UTC") {
    val offset = DateTimeZone.getDefault().getOffset(System.currentTimeMillis()) / 1000
    val FIVE_HOURS = 5 * 60 * 60
    startTS -= offset
    startTS += FIVE_HOURS
    if (endTS != 0) {
        endTS -= offset
        endTS += FIVE_HOURS
    }
}
// 2
```

After this, if the user's timezone is further ahead than GMT+5 i.e. `FIVE_HOURS - offset < 0`, the timestamp is now some time in the previous UTC day.

([CalDAVHandler.kt L261](https://github.com/SimpleMobileTools/Simple-Calendar/blob/master/app/src/main/kotlin/com/simplemobiletools/calendar/helpers/CalDAVHandler.kt#L261))

```kt
if (event.getIsAllDay() && endTS > startTS) {
    event.startTS = Formatter.getShiftedImportTimestamp(event.startTS)
    event.endTS = Formatter.getShiftedImportTimestamp(event.endTS)
    event.endTS -= DAY
}
// 3
```

`getShiftedImportTimestamp` takes the UTC timestamp and reinterprets it as local time, and then sets the clock time to 0500 local time, then converts it back to a Unix timestamp (i.e. UTC).

The problem is that if the local TZ is ahead of GMT+5, the UTC timestamp and consequently the reinterpreted local timestamp end up one day earlier.

Here's some real values for an all-day event on 11 April 2018. My timezone is GMT+8.

* @ 1: 1523404800: 2018-04-11 00:00:00 UTC = 2018-04-11 08:00:00 GMT+8
[...]

PR #369, opened 2018-02-18
Fix all-day CalDAV event import for local TZ > GMT+5

angelsl:
I'm not sure if this will affect whatever bug was being fixed in 91d5c3dd00e16320ede02b71ca07eb7acd1f9fa7.


tibbi:
so with the old version if I set the timezone to more than +5 and use Google sync, I can reproduce some issue? All-day sync is apparently still having issues as Google, Yahoo, NextCloud, Microsoft etc do things a bit differently and whatever I fix, something else breaks.

angelsl:
> so with the old version if I set the timezone to more than +5 and use Google sync, I can reproduce some issue?

Yeah (see #368, I debugged the issue there).

> All-day sync is apparently still having issues as Google, Yahoo, NextCloud, Microsoft etc do things a bit differently and whatever I fix, something else breaks.

Hmm, do you have some examples of how each sync adapter represents all-day events? Maybe we can work this out.

tibbi:
I havent really written down how does it look like at which service, I will try your fix with some of them. It would be good if it worked as you removed that ugly part.

tibbi:
Seems to be working fine in my cases, we will see real results once it gets published. Thanks

PR #388, opened 2018-03-02
add default public notification

### LINKED ISSUE #370: Don't show sensitive information on lockscreen notification if turned on.
I have turned on to *NOT* show any sensitive information in notifications on lockscreen. I would expect that the calendar is only showing that I have an event at a given time but not the information what is is about on the lockscreen.

SylvainGirod:
This is a quick fix for the mentioned issue.
Fixes #370


tibbi:
this is a bit too much code for fixing a single flag, it needs a simpler solution

SylvainGirod:
There is another solution, but the displayed message will be generic, because set by android

tibbi:
I will research it better when I get to it, but this isnt really a solution I like. Thanks though

PR #402, opened 2018-03-20
 Improve notification content by displaying date if useful

fraang:


tibbi:
I like the idea, thanks. It needs a couple tweaks, but it will be the easiest if I do it myself :) Normally the string would have to be added in all strings files, even if obviously not translated, but I will move it in Commons anyway as it is a pretty common word and I might use it in other apps too.

fraang:
Thanks. :-) I would like to continue contributing to the simple mobile tools. How is the development coordinated? Is there any chat or messaging service?

tibbi:
not really, since Im the only developer :D Just pick some issue and let me know that you are going to work on it, just to be sure you do something that makes sense and I want it that way. Thanks

PR #429, opened 2018-04-10
Add a public version of the notification to make it possible to optio…

fraang:
…nally hide sensitive information

Also don't override the visibility to respect the user's notification
settings

Closes Issue #370

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/extensions/Context.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/extensions/Context.kt
@@ -167,7 +167,7 @@ fun Context.notifyEvent(event: Event) {
 }
 
 @SuppressLint("NewApi")
-fun Context.getNotification(pendingIntent: PendingIntent, event: Event, content: String): Notification {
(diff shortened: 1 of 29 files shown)


app/src/main/kotlin/com/simplemobiletools/calendar/extensions/Context.kt
```diff
@@ -184,9 +184,17 @@ fun Context.getNotification(pendingIntent: PendingIntent, event: Event, content:
     val soundUri = Uri.parse(config.reminderSoundUri)
     grantReadUriPermission(config.reminderSoundUri)
 
+    val contentTitle = when (publicVersion) {
+        false -> event.title
+        true -> resources.getString(R.string.app_name)
+    }
```
tibbi: oh gosh, rewrite this to "val contentTitle = if (publicVersion) resources.getString(R.string.app_name) else event.title" please

tibbi:
thanks, I will check the actual functionality later, added a note for now

app/src/main/kotlin/com/simplemobiletools/calendar/extensions/Context.kt
fraang: no problem

PR #502, opened 2018-05-23
Update strings.xml

FatihFIRINCI:
Sorry for typo


tibbi:
no problem, just keep the apostrophes escaped in the whole file please, else it gives me an error in android studio

FatihFIRINCI:
In this case, just playstore description should be fixed. Seems to be "\" character in the description.

PR #579, opened 2018-08-12
FAQ - New and corredted translations

10cents:
A question:  
Redundant "too" in "affected by the system too"?


tibbi:
its ok, no big deal :)

PR #585, opened 2018-08-15
german Strings updated

michaelof:
Hi Tibor,

tried to find all (new) German strings, currently in English. All I've found translated in this patch/pull request.
"reminder audio streams" somewhat hard to translate, I've followed my Android 6.0 German settings names  at "settings" / "Sound&notification" / "Volumes". From an end user perspective, audio stream "system" is called "media volume", also in English. Changed FAQ also.

CAN'T find keys for:

- "Purchase Simple Thank You"
- "Snooze ..." (Not sure if this might be even "snooze" in German, as most are used to this English wird, German verb is "schlummern")
- "Customize widget colors"

They seems to be inherited. Or maybe (e.g. settings_snooze_time_label), not ready to be localized right now?

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -159,11 +159,11 @@
     <string name="loop_reminders">Erinnerungen wiederholen bis sie verworfen werden</string>
     <string name="dim_past_events">Vergangene Termine ausgrauen</string>
     <string name="events">Termine</string>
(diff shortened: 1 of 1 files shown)


tibbi:
hey, there are many strings in the Commons library, which is reused by multiple apps. You can find it at https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/res/values-de/strings.xml

PR #586, opened 2018-08-16
Changed Mariä to Maria

Krzmbrzl:
Afaik it is way more common to call it "Maria Himmelfahrt" instead of "Mariä Himmelfahrt"


tibbi:
well I cannot decide, thanks :)

PR #618, opened 2018-09-12
Add Malaysian holidays

youdly:
[Source](https://calendar.google.com/calendar/ical/en.malaysia%23holiday%40group.v.calendar.google.com/public/basic.ics). Include regional holidays.


youdly:
If you don't mind the source, I can add/update of other countries as well.

tibbi:
you havent added the holidays file itself. Also, please use only tags mentioned at https://github.com/SimpleMobileTools/Simple-Calendar/pull/619

youdly:
I added `app/src/main/assets/malaysia.ics` which is the holidays file? I'm cleaning up the file.

tibbi:
yeah that looks good, thanks. I guess you tried importing the .ics file and it worked fine, right?

youdly:
Tried importing this and https://github.com/SimpleMobileTools/Simple-Calendar/pull/619 on v4.1.3. No issue. All 2019 holidays are shown. Constant-date holidays are shown in 2020.

PR #635, opened 2018-09-23
Updated German translation.

Atalanttore:
Translated newly added strings to german.

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -164,10 +164,10 @@
     <string name="alarm_stream">Wecker</string>
     <string name="notification_stream">Benachrichtigung</string>
     <string name="ring_stream">Klingelton</string>
-    <string name="use_last_event_reminders">Use the last events\' reminders as the default for new event</string>
-    <string name="default_reminder_1">Default reminder 1</string>
-    <string name="default_reminder_2">Default reminder 2</string>
-    <string name="default_reminder_3">Default reminder 3</string>
+    <string name="use_last_event_reminders">Erinnerungen der letzten Termine als Standard für neuen Termin verwenden</string>
+    <string name="default_reminder_1">Standarderinnerung 1</string>
+    <string name="default_reminder_2">Standarderinnerung 2</string>
+    <string name="default_reminder_3">Standarderinnerung 3</string>
 
     <!-- CalDAV sync -->
     <string name="caldav">CalDAV</string>


tibbi:
@Atalanttore a guy reported that the app names like `Schlichte Gallery` or so sound bad, like something stupid or crappy. He says that `Einfache Gallery` sounds better. Any ideas on that? I cannot really device. Thanks

tibbi:
@be-we so you advise using Einfach, right?

be-we:
In my opinion yes.

ouzowtf:
But Einfach could also mean not very feature rich or not very polished.
And these are things I wouldn't use to describe your very nice app :grinning:

tibbi:
well ye, initially I wanted to keep it low on features, but it somehow got out of control :)

tibbi:
So, another guy reported that the translation of "Import contact birthdays" could be also misleading. Currently it is "Kontaktgeburtstage hinzufügen", he suggests "Kontakt zu Jahrestagen hinzufügen". Does it make sense?

ouzowtf:
No, it does not make any sense. This would be like "add contact to anniversary".

Atalanttore:
"Import contact birthdays" translates as "Geburtstage von Kontakten importieren".

tibbi:
ye I mixed up birthdays and anniversaries now, but ignore that. So which one is better,the current  "Kontakt zu Jahrestagen hinzufügen" or a new suggestion "Jahrestage der Kontakte hinzufügen"?

ouzowtf:
Generally "Jahrestage der Kontakte hinzufügen" is better, but it should be similar with the other options for holidays and .ics-import.
Do you want to use "import" or "add"?

import:
"Feiertage importieren"
"Geburtstage der Kontakte importieren"
"Jahrestage der Kontakte importieren"
"Termine der .ics-Datei importieren"

add:
"Feiertage hinzufügen"
"Geburtstage der Kontakte hinzufügen"
"Jahrestage der Kontakte hinzufügen"
"Termine der .ics-Datei hinzufügen"

While we're at it:

Your "Filter events by type" should be "Termine nach Typ filtern" instead of "Termine nach Typ sortieren" in German, because there is no sorting.

tibbi:
holidays, birthdays and anniversaries should be "Add", just .ics files have "Import" also in english, so lets stick to that. Fix the filtering string too please with a new pull request. Thanks

PR #669, opened 2018-10-22
Added setting to let user choose his favorite default view

### LINKED ISSUE #424: [FR] Make default view selectable
Hi,

I would find it really helpful to make the default view selectable in the settings.

Currently: when I press the top bar of my widget I end up in the daily view.

What I would like to see: monthly view.

Best regards and many thanks for the great apps (will donate soon) ;)

knusprjg:
- Added "default view" option to settings that allows user to choose which calender view is launched when pressing at the top bar of the events list widget (closes #424 )
- Removed "OPEN_MONTH" extra boolean for selected view and replaced it with more versatile integer to achieve more flexibility
- Added two more strings so far only in english and german


knusprjg:
I was hoping you had a automized way to transfer the strings to all languages ;) I tried the translation editor but it appears to be kind of useless?! Now I did it using the search function.

tibbi:
Im doing it manually too, I couldnt find any better way of adding/editing/removing strings :/ That built in editor sucks.

tibbi:
looks good, Ill do just some smaller tweaks. Thanks

knusprjg:
Great, I thank you! :)

Issue #664, opened 2018-10-16
Feature Req:  pull to refresh gesture


alaricljs:
For calendar views that support it, it would be nice to have the pull gesture (aka swipe down) trigger a refresh of CalDAV calendars.

azisuazusa:
Hi @tibbi, can I work on this issue?

tibbi:
ye

PR #670, opened 2018-10-23
#664 pull to refresh feature

azisuazusa:
#664 pull to refresh feature


tibbi:
alright looks good, I just did some minor updates. Thanks

tibbi:
I actually made it optional, disabled by default. It made scrolling quite glitchy and people wont want to use it often anyway, there are many syncs in the background.

PR #684, opened 2018-10-30
Fix End Date of All-Day Events

lkiesow:
Setting a single day, all-day event, Simple Calendar will set both the
start and end date of the event to the specific date which is incorrect
according to [rfc5545]. The end date is non-inclusive and must hence be
set to the following day:

> The "DTEND" property for a "VEVENT" calendar component specifies the
> non-inclusive end of the event.

This causes weird effects with other calendar apps with e.g. Khal
complaining about the invalid ical or Sogo displaying the event the day
before it should happen.

There is a special exception for one-day, all day events in the
[rfc5545], but that would require no DTEND to be set at all:

> For cases where a "VEVENT" calendar component specifies a "DTSTART"
> property with a DATE value type but no "DTEND" nor "DURATION"
> property, the event's duration is taken to be one day.

Note that this is already correctly handled for multi day events in the
CalDAV export and for all types of events in the .ics export.

[rfc5545] https://tools.ietf.org/html/rfc5545#section-3.6.1

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/helpers/CalDAVHandler.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/helpers/CalDAVHandler.kt
@@ -325,7 +325,7 @@ class CalDAVHandler(val context: Context) {
                 put(CalendarContract.Events.RRULE, repeatRule)
             }
 
-            if (event.getIsAllDay() && event.endTS > event.startTS)
+            if (event.getIsAllDay())
                 event.endTS += DAY
 
             if (event.repeatInterval > 0) {


tibbi:
thanks, but that check is needed for some more popular calendars than Khal. Sadly those standards are just rough rules, apps implement them in different ways.

lkiesow:
Can I urge you to rethink this decision?
I've checked some popular calendars and couldn't find this behavior anywhere else:

- Google Calendar will set the end date one day after the start date which would be consistent with the behavior of this patch
- ownCloud's calendar does the same
- Sogo's calendar will do the same and additionally fail to parse your invalid event correctly, causing it to be displayed the day before

Hence, the apps you talk about would be incompatible to all of these calendars.

There are also a few other good arguments to be made:
- This code is behaving differently compared to your own ics export code which will set the end date a day after the start date
- This check does not work for all single-day events, making the app inconsistent in that it sometimes generates “correct” iCalendar events (create event → change end time → then make it an all-day event)
- Without this check, the app generates the more general form of events with start and end date explicitly set, avoiding the special cases (e.g. no end date or duration) which calendars should always be able to parse properly or else they would also handle multi-day, all-day events incorrectly.

And finally, it's always good to not deliberately ignore the standard. I'm absolutely certain that there is some code out there doing that (there always is ;-)) but given the choice and knowledge, it's usually better to stick to the standard and not behave more or less randomaly different bacause that makes interaction between applications even harder.

PS: Out of interest, which calendars cannot handle this?

tibbi:
I dont remember now which calendars had this specific issue, but it is there for a reason. It seems to work fine now in all cases, I dont remember people reporting offset issues. I dont see any Sogo or Khal related android apps.

PR #750, opened 2019-01-09
Update strings.xml

smarquespt:


tibbi:
@smarquespt please make the short description shorter than 80 chars

Issue #751, opened 2019-01-10
RTL support


karmac2015:
Hello, I like your very nice apps, But I think if you support RTL layout it should be better.

jtj9817:
Hey @tibbi , can I take this issue? I will probably need guidance on implementing the RTL support since it will require changing quite a number of layouts but I'll try to work on what I can.

karmac2015:
Hello, @jtj9817 I have finished  modifying the project Simple-Contacts to support RTL, take a look at the changes I have made [here](https://github.com/karmac2015/Simple-Contacts.git)

jtj9817:
Hey @karmac2015 , thanks for the reference code. I'm sure I will find it helpful and I'll take a look at it.

schapira:
can I help in that? what can I do?

tibbi:
only developers can help, with adjusting the appropriate code and layouts

schapira:
I am a junior developer...

On May 20, 2019 11:25:41 AM GMT+03:00, Tibor Kaputa <notifications@github.com> wrote:
>only developers can help, with adjusting the appropriate code and
>layouts
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/SimpleMobileTools/Simple-Calendar/issues/751#issuecomment-493887690

תודה
אור שפירא

tibbi:
well so test the app and see what xml changes need to be done, so that it works as expected

PR #762, opened 2019-01-25
Phase One for RTL Support #751

jtj9817:
Please review this PR for issue #751 and if there are any changes that need to be made please comment on them.


app/release/output.json
```diff
@@ -0,0 +1 @@
+[{"outputType":{"type":"APK"},"apkInfo":{"type":"MAIN","splits":[],"versionCode":141,"versionName":"6.2.1","enabled":true,"outputFile":"calendar-release.apk","fullName":"release","baseName":"release"},"path":"calendar-release.apk","properties":{}}]
```
tibbi: remove these 3 files from the pull request

jtj9817: Just finished deleting them. Any ideas on why the /release folder was present?

tibbi: not sure how come it doesnt show to me either actually, since it isnt added to .gitignore. Anyway before doing a commit always check what files are modified. Looks good now, thanks.

PR #779, opened 2019-02-09
fix #772, optimize drawing of events in MonthView

### LINKED ISSUE #772: Overlapping events in monthly view partially invisible
In the monthly view some overlapping events don't show corectly.

Setup:

- Event "Foo" from October 18 to 22.
- Event "Boo" from October 19 to 23.

Like that the events show correctly. But if "Boo" is set to October 19 to 24, suddenly the part of "Boo" from October 19 to 20 becomes invisible. This also happens when the end-date is later than October 24.

Unfortunately this is a big bug in the monthly view (almost the only one I use), as it means I can not fully trust the dates in my calendar, without clicking on every single event!

I have tried to reproduce the bug at different places in the calendar, and it happens other places too. I am not completely sure what triggers it though.

cbertram:
There's an informal proof that the algorithm works in the commit message. (I can elaborate if you want.)

The algorithm has runtime O(n*log(n)), so it works nicely even with a ton of events in the month.

No problem if you want to throw this away, and write your own fix. I had fun making it :)

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/views/MonthView.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/views/MonthView.kt
@@ -24,6 +24,7 @@ import com.simplemobiletools.commons.extensions.getContrastColor
 import com.simplemobiletools.commons.extensions.moveLastItemToFront
 import org.joda.time.DateTime
 import org.joda.time.Days
+import java.util.TreeSet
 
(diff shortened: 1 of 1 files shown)


tibbi:
it doesnt work properly, stucking at the while() loop already with a few events. Thanks though

cbertram:
I'm sorry, it seems I pushed the wrong version.. At line 122 it should be `-1000000000`, not `1000000000`!

PR #780, opened 2019-02-09
Add default start time and default duration for events in settings

### LINKED ISSUE #696: request: default time for events
Hi, one thing I really miss on simple calendar ist a default time for events. When creating a new event start and end time is always the same time, which is not useful.
Would it possible to change that

NWuensche:
Add settings for default start time and duration when adding an event. Close #696 .


tibbi:
ok thanks, Ill do some more tweaks myself

PR #781, opened 2019-02-13
Update strings.xml

spkprs:


tibbi:
I copied the values manually, as Ive just added another new string which is causing conflicts and I cannot accept your edit without modifying. Thanks

PR #840, opened 2019-03-28
Add Taiwan holidays

abc0922001:
follow https://calendar.google.com/calendar/embed?src=zh-tw.taiwan%23holiday%40group.v.calendar.google.com&ctz=Asia%2FTaipei


tibbi:
nice, thanks. I will just filter out some invalid tags

PR #847, opened 2019-04-03
Added permission request for READ_CALENDAR and WRITE_CALENDAR

AkashPratik:


tibbi:
it is not required to have those permissions

PR #862, opened 2019-04-16
fixes #860: whole day event do not ignore time

### LINKED ISSUE #860: whole day event do not ignore time
Reproduce:
1. create event (from 2019-05-01 16:00 to 2019-05-03 15:00)
2. change to whole day mode.
3. save event
4. modify date to single day event (from 2019-05-01 to 2019-05-01)

Issue:
View shows error because the to date is before the from date. This is because the date from and date to fields do not ignore the time that is not used in a whole day event.

Possible Solution:
Set the time in whole day events to 00:00 in from date and to date.

k0ur0x:


tibbi:
thanks, I will accept it and do a couple changes

PR #879, opened 2019-05-06
fix: content goes behind the naviagtion bar

jaypratapsingh:


tibbi:
that isnt related to the issue at all, actually number showing behind navigation bar isnt even an issue.

Issue #843, opened 2019-03-31
Slightly rounded corners instead oft sharp edges


daehruoydeef:
Would be really nice if the event inside the calendar would get slightly rounded corners instead of this sharp edges. This would go way better with the new Android Design and would look cleaner and more modern.

rjain90:
Hi, Can I work on this?

tibbi:
sure

rjain90:
Fixed in #884

PR #884, opened 2019-05-09
event rounded corners

rjain90:
Fix for #843

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -4,9 +4,9 @@ apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlin-kapt'
 apply plugin: 'de.timfreiheit.resourceplaceholders'
 
-def keystorePropertiesFile = rootProject.file("keystore.properties")
-def keystoreProperties = new Properties()
-keystoreProperties.load(new FileInputStream(keystorePropertiesFile))
+//def keystorePropertiesFile = rootProject.file("keystore.properties")
+//def keystoreProperties = new Properties()
+//keystoreProperties.load(new FileInputStream(keystorePropertiesFile))
 
 android {
     compileSdkVersion 28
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -4,9 +4,9 @@ apply plugin: 'kotlin-android-extensions'
 apply plugin: 'kotlin-kapt'
 apply plugin: 'de.timfreiheit.resourceplaceholders'
 
-def keystorePropertiesFile = rootProject.file("keystore.properties")
```
tibbi: even if you did these changes locally, you definitely shouldnt have commited them. Revert it please.

app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/SimpleActivity.kt
```diff
@@ -68,4 +68,6 @@ open class SimpleActivity : BaseSimpleActivity() {
     private fun unregisterObserver() {
         contentResolver.unregisterContentObserver(calDAVSyncObserver)
     }
+
```
tibbi: these extra lines also arent needed

tibbi:
creating a new pull request might be the easiest way of getting rid of those unwanted changes. Thanks

app/build.gradle
rjain90: This file is not in the master branch. In order to build the code, I'm assuming everyone has to comment these lines.

tibbi: you shouldnt care about everyone else, only about the issue you are fixing right at the moment

rjain90: - Everyone includes me. How would I build the code successfully without the "keystore.properties" file?

tibbi: you can change any code locally, but you shouldnt push it to the master branch

PR #904, opened 2019-05-28
Update strings.xml

Anke:
I translated the following lines to German:
13
14
88
95
105
106
110
185
186
187
188
189
190
191
205
241
243
245
247
249
255-272
274
276

The only German terms missing here, are in lines 107-109, because it is unclear in which context the terms "Going", "Not going" and "Maybe going" are being used. Since that part is about attendees, the term "going" could mean something like "possible" or "leaving". Translating the wrong meaning could end up sounding rather ridiculous.

Anka

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -103,10 +103,10 @@
 
     <!-- Event attendees -->
     <string name="add_another_attendee">Teilnehmer hinzufügen</string>
-    <string name="my_status">Mein Status:</string>
-    <string name="going">Going</string>
-    <string name="not_going">Not going</string>
-    <string name="maybe_going">Maybe going</string>
+    <string name="my_status">Mein Teilnahmestatus:</string>
+    <string name="going">Ja</string>
+    <string name="not_going">Nein</string>
+    <string name="maybe_going">Vielleicht</string>
     <string name="invited">Eingeladen</string>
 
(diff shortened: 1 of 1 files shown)


tibbi:
those strings are in the context of attendance indeed, so "going" means that the person will attend the event. Thanks

Anke:
Sorry for asking again. I need to know how to conjugate the verbs and I can't reproduce this, as there are no attendance options in my calendar.

The string names in the attendee section are "add_another_attendee", then "my_status". Are the following strings "going/not_going/maybe_going" referring to **another invited attendee** or to **my status**? 

In other words: Does "going" mean, I am attending or he/she is attending? Or are those variables actually applying to both cases?

tibbi:
Attendance is visible only at CalDAV synced events once you add an attendee. Those textual statuses are related to myself only, the attendance of others is marked with a colored icon. Im attaching an image, the other person would have a status color at the bottom right corner of his image, in case he already set any satus.
Screenshot_20190530-170934

Anke:
I see. THX.
Isn't this usually called "Online-Status"? Like "online", "offline" or "online status hidden"?

tibbi:
it has nothing to do with being online or not, but joining the meeting. Just check Google Calendar how is it translated there

Anke:
Okay, I completed the translation. Thank you for your patient help.

PR #930, opened 2019-07-08
Update strings.xml

AO-LocLab:
Typographical changes. Changed translated app name from "Agenda simple" to "Simple Agenda" to be consistent with other Simple app names and closer to the English name. Changed all "calendrier" to "agenda". Some string reviews.


tibbi:
’ doesn't need a backward slash in front of it, only ' does. I mean the one on line 18 that you removed had to have it, but the ones you added on 19 don't. You probably dont have to change it now, just saying for the future. Thanks

PR #977, opened 2019-09-15
Always shift end date by one day for all-day events

### LINKED ISSUE #976: Possibly wrong DTEND value für single-day event
When creating an All-day event for a single day Simple Calendar (version 6.6.0) sets DTSTART and DTEND to the same value: 
```
DTSTART;VALUE=DATE:20191001
DTEND;VALUE=DATE:20191001
```

My desktop calendar program (khal) warns about these kinds of events:
> The event's end date property contains the same value as the start date, which is invalid as per RFC 5545. Khal will assume this is meant to be single-day event on 2019-10-01.

I guess this is correct since [RFC 5545](https://tools.ietf.org/html/rfc5545#section-3.8.2.2) states
> its [DTEND] value MUST be later in time than the value of the "DTSTART" property.

I also confirmed that the default AOSP calendar app would create a single-day event with the following values:
```
DTSTART;VALUE=DATE:20191001
DTEND;VALUE=DATE:20191002
```

ddast:
Without this change the end date is only shifted by one day when it is already ahead of the start date.  However, when an all-day event for a single day is created where the start and end times are equal (before selecting all-day) then the end date is not shifted by one day.  This commit changes the behaviour so that the end date is also shifted by one day in this case.

I checked that this fixes the problem described in #976.  Note that I know almost nothing about CalDAV or the internals of Simple Calendar so I might be doing something stupid.

Fixes #976.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/CalDAVHelper.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/CalDAVHelper.kt
@@ -409,7 +409,7 @@ class CalDAVHelper(val context: Context) {
                 put(CalendarContract.Events.RRULE, repeatRule)
             }
 
-            if (event.getIsAllDay() && event.endTS > event.startTS)
+            if (event.getIsAllDay() && event.endTS >= event.startTS)
                 event.endTS += DAY
 
             if (event.repeatInterval > 0) {


tibbi:
will take a look at it and test it a bit later, not working on the calendar right now. Thanks

tibbi:
it didnt break anything at syncing via Google, so if it fixed your syncing, guess it is good to go. Thakns

PR #1020, opened 2019-11-24
Update norway.ics

FTno:
Extended norwegian holiday file. Tested to work fine.

Just a question/idea - Could Show/Hide specific country holidays be implemented?
The reason: Sometimes when being in other countries, it could be useful to have the countries holidays in the calendar temporarily.
Now everything must be removed and then you have to add again, that's tedious if doing this rather frequently.


tibbi:
it isnt really on the roadmap for now, might be added in some form later. You can just delete the whole Holidays event type from the app settings, then reimport the wanted holiday till it gets changed. Thanks

PR #1075, opened 2020-02-02
Add Bengali translation

nowshad-hasan:


tibbi:
cool, thanks. I dont see such a language on Google Play yet, guess it isnt supported so I cannot apply the app descriptions there.

nowshad-hasan:
I think it's supported. I'm using Android 10 on my Nokia 7 Plus which is stock android. It's very easy to change the language from settings to Bengali. You can try it on your phone too.

tibbi:
BN belongs to Bangla on Google Play. Is that it?

nowshad-hasan:
Yap, Bangla and Bengali are the same.

tibbi:
ok thanks, I applied it

PR #1086, opened 2020-03-06
Update Turkish strings.xml

FatihFIRINCI:


tibbi:
thanks, but the whole app description is actually untranslated together with the short description and title, not just these 2 lines :)

FatihFIRINCI:
I did not see, sorry :)

PR #1117, opened 2020-04-14
changed German and French translations

Honk2:
German: removed inappropriate hyphens and added missing translations
French: changed unescaped single quotes to apostrophes


tibbi:
nice, thanks

PR #1126, opened 2020-04-19
improve week number display on month view

Honk2:
* suppress colon of week number if grid active
* dim week number like week day if week does not belong to displayed month


tibbi:
I dont really like those changes, I prefer the current way. Thanks though

Honk2:
ok, regarding the week number color, it is a matter of taste.
But I think the colon is somehow too much if the number is already separated by the grid.
I would by happy if you would reconsider the colon.

Issue #969, opened 2019-08-27
German translation


hurzl:
In German it is "7. September", not "September 7".

khiarihamdi:
i can work on this if someone would provide me with some details.

tibbi:
duplicate of #1055

PR #1127, opened 2020-04-19
make getDateFromCode and getDayTitle locale aware

Honk2:
fixes SimpleMobileTools/Simple-Calendar#969

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Formatter.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Formatter.kt
@@ -8,6 +8,7 @@ import org.joda.time.DateTime
 import org.joda.time.DateTimeZone
 import org.joda.time.LocalDate
 import org.joda.time.format.DateTimeFormat
+import java.util.Locale
 
 object Formatter {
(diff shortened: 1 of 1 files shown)


tibbi:
that doesnt solve the situation, the app needs a "Change date format", like some apps have. This would work really weird.

Honk2:
I am not sure what a "change date format" option should look like.
Usually a user likes to see dates and numbers formatted to a specific locale. Typically the one your are familiar with. Therefore I think using locale functions is a good approach. 
Of cause it could be a nice feature to add an option to select whether to use the system default locale or a specific one.
Would this be acceptable and solve the situation?

tibbi:
you can see the date format picker for example in the gallery. Also with your solution the user would see 2020 at every date, which would be really annoying and ugly.

Honk2:
Ah, just checked the date format picker and got your point.

Still I would suggest to use locale functions and let the user select the short or long version of the selected locale. This would include the choices in the format picker and prevent requests for more formats, because they are all included.

The missing year is actually why I still stick to the default (Lineage) calender. If this is annoying and ugly is a matter of taste. Some like to hide any redundant data, others like me prefer consistent layouts as in the default calender.

Anyway I see that a comprehensive date format picker with preferences for the different views is the key and a real advantage for your calender.

Issue #1096, opened 2020-03-13
Parse error when BYMONTHDAY is a list


PaulKiddle:
Exception is being thrown on this line https://github.com/SimpleMobileTools/Simple-Calendar/blob/59f94947534a10ef2f913016b941b8df4f9a34ef/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Parser.kt#L57

When BYMONTHDAY is a comma-separated list of numbers

I think this is preventing other events from syncing. 

I don't know Kotlin but a quick search tells me you should be able to switch that `toInt` to `toIntOrNull`

_Originally posted by @PaulKiddle in https://github.com/SimpleMobileTools/Simple-Calendar/issues/1095#issuecomment-598724585_

saunaklogan:
Thought of working on this, can I get some more information on how to reproduce this issue ?

PaulKiddle:
1. On the server add this event to the calendar: https://github.com/SimpleMobileTools/Simple-Calendar/files/4325553/bymonthday.ics.txt
2. Try to sync client
3. Look out for java parsing error

On success there should be a monthly repeating "WRAG Committee" event in the calendar - though I'm not sure which day(s) it will repeat on due to the issue in #1095

pcsaunak:
@PaulKiddle Thanks for the details. 
I was going through an article here 
[https://www.kanzaki.com/docs/ical/recur.html](url) and it seems that BYMONTHDAY will have comma separated values and its possible values could range from 1 to 31 or -31 to -1. So we should do an individual check for each of the values. For example instead of 
```
else if (key == BYMONTHDAY && value.toInt() == -1) { 
}
```
we should break this up into two parts:

```
else if (key == BYMONTHDAY){
     val byMonthDayValues = value.split(",")
     ....... 
     .......
}
```

@tibbi what are your thoughts here ??

tibbi:
well test some fix, and if it works, create a pull request

pcsaunak:
Got a good solution and it created the recurring event 
<img width="163" alt="image" src="https://user-images.githubusercontent.com/23631671/80985259-99f74800-8e4c-11ea-8d10-6527d6d1589f.png">

pcsaunak:
@tibbi  this issue is fixed. We can close this I guess. Anything else needs to be done ?

tibbi:
well ye I can close this, the parent issue is #1095. -1 is just one of the cases, there are many others.

PR #1132, opened 2020-05-04
Fix for parsing .ics file when the key BYMONTHDAY is present in the .…

pcsaunak:
…ics file.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Parser.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Parser.kt
@@ -60,8 +60,11 @@ class Parser {
                 } else if (repeatInterval.isXMonthlyRepetition() || repeatInterval.isXYearlyRepetition()) {
                     repeatRule = if (value.startsWith("-1")) REPEAT_ORDER_WEEKDAY_USE_LAST else REPEAT_ORDER_WEEKDAY
                 }
-            } else if (key == BYMONTHDAY && value.toInt() == -1) {
-                repeatRule = REPEAT_LAST_DAY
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Parser.kt
```diff
@@ -60,8 +60,11 @@ class Parser {
                 } else if (repeatInterval.isXMonthlyRepetition() || repeatInterval.isXYearlyRepetition()) {
                     repeatRule = if (value.startsWith("-1")) REPEAT_ORDER_WEEKDAY_USE_LAST else REPEAT_ORDER_WEEKDAY
                 }
-            } else if (key == BYMONTHDAY && value.toInt() == -1) {
-                repeatRule = REPEAT_LAST_DAY
+            } else if (key == BYMONTHDAY) {
+                val byMonthDayValueArray = value.split(",")
+                val intFlag = byMonthDayValueArray.find { it.toInt() == -1 }
```
tibbi, line 65: whats byMonthDayValueArray?

PaulKiddle:
Shouldn't that be `if (intFlag != null)` ?
Only want to set the `REPEAT_LAST_DAY` rule if `-1` is in the list (in the example from #1096, the bymonthday list does not contain `-1`).

pcsaunak:
> Shouldn't that be `if (intFlag != null)` ?
> Only want to set the `REPEAT_LAST_DAY` rule if `-1` is in the list (in the example from #1096, the bymonthday list does not contain `-1`).

Yup you are correct. I just checked the old code it was checking -1. I will cancel this and create a new one.

PR #1133, opened 2020-05-04
Parsing of .ics file for key BYMONTHDAY

pcsaunak:
Recreated the pull request as there was a wrong check in the previous one

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Parser.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Parser.kt
@@ -60,8 +60,11 @@ class Parser {
                 } else if (repeatInterval.isXMonthlyRepetition() || repeatInterval.isXYearlyRepetition()) {
                     repeatRule = if (value.startsWith("-1")) REPEAT_ORDER_WEEKDAY_USE_LAST else REPEAT_ORDER_WEEKDAY
                 }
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Parser.kt
```diff
@@ -60,8 +60,11 @@ class Parser {
                 } else if (repeatInterval.isXMonthlyRepetition() || repeatInterval.isXYearlyRepetition()) {
                     repeatRule = if (value.startsWith("-1")) REPEAT_ORDER_WEEKDAY_USE_LAST else REPEAT_ORDER_WEEKDAY
                 }
-            } else if (key == BYMONTHDAY && value.toInt() == -1) {
-                repeatRule = REPEAT_LAST_DAY
+            } else if (key == BYMONTHDAY) {
+                val byMonthDayValueArray = value.split(",")
+                val intFlag = byMonthDayValueArray.find { it.toInt() == -1 }
```
tibbi, line 65: instead of .find try using .any, should be easier to read. Then you wont need the intFlag variable

tibbi:
ok I tweaked and accepted it without really testing, should work well. That repetition rule is really rare so I dont think it will be used often. It is a more advanced one, using it with REPEAT_LAST_DAY is just one of the cases

Issue #1055, opened 2020-01-05
Can't find the option to change date format


Obigre:
Hello !
After activating the widget i noticed the date format there was not what i wanted (month/day number instead of the opposite). I searched the options to find where i could find that but didn't see anything.
Can you help me ? or if it's not a feature, can we make it into one ?

tibbi:
that ticket was about the time format (24 hours format or not), not the date format. Adding date format customization is on the roadmap.

Obigre:
Thank you for your answer :)

im-indrajeet:
Thanks, Can you tell me in broad sense what's the way of doing it.
This is my first time to contribute to any open source I really like to know and contribute.

ldmpub:
I confirm the above requested enhancement => see below screenshot of the date displayed in "Events List View".
In french, it should be "14 avril (Mer.)" or even other formats like "Mer. 14 avril".

![Date Display](https://user-images.githubusercontent.com/1452146/114787114-87390300-9d7f-11eb-8089-058fde5a8241.jpg)

ldmpub:
Other sample in the widget: date with year should be something like "Mercredi 14 avril 2021" (in french) 👍 
![Widget](https://user-images.githubusercontent.com/1452146/114787843-b13ef500-9d80-11eb-8137-972851368069.jpg)

hurzl:
Can this be so complicated? Can't the date format simply be requested from the system, like https://stackoverflow.com/a/18982842 says:

```
Format dateFormat = android.text.format.DateFormat.getDateFormat(getApplicationContext());
String pattern = ((SimpleDateFormat) dateFormat).toLocalizedPattern();
```

ltGuillaume:
Honestly, this should be highest priority.

hurzl:
there is also `android.text.format.DateFormat.getBestDateTimePattern`:
"One difference is that order is irrelevant. For example, "MMMMd" will return "MMMM d" in the en_US locale, but "d. MMMM" in the de_CH locale."

emes81:
This issue was opened almost three years ago - is a fix for this expected? A calendar app needs to be able to handle dates properly and this should be the number one development priority for Simple Calendar. Until this is fixed, the app is not usable.

This is also emphatically not a feature request; it is a bug.

Cwpute:
[...]

PR #1147, opened 2020-05-24
added DateformatComposerDialog

Honk2:
added a date format composer dialog
added option for event list section dateformat using DateformatComposer
added option for event list section header size

could be used for #1055 and all other locations where a date should be displayed and implements #969


tibbi:
create 1 pull request per issue please, you are mixing totally unrelated things. And many of those strings are already in Commons, do not duplicate them. Anyway, I dont want such an advanced date format picker, it is seriously overengineered. Thanks though

Honk2:
Sorry for not looking in Commons for the strings.
What kind of format picker are you looking for? A predefined list?
That could result in several lists, depending on where the date is to be displayed. My intend was to create a format picker, that could be used for every date displayed. Because there are several issues related to date formats. And what about other locales and preferences? Sure it is may not be easy for everyone to use but maybe as an advanced option? What do you think/want? Which dates should get a format picker? Please let me know.

tibbi:
the one used in some other apps from our suite. Like the SMS Messenger.

PR #1154, opened 2020-06-05
closes #1084

### LINKED ISSUE #1084: Add a warning when closing the Edit screen with unsaved changes

fnrngg:
Needs translations, dialog strings currently marked as non-translatable


fnrngg:
#1084

app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/EventActivity.kt
```diff
@@ -240,11 +240,27 @@ class EventActivity : SimpleActivity() {
...
         }
         return true
     }
 
+    override fun onBackPressed() {
+        CloseEditingDialog(this) {
+            ensureBackgroundThread {
+                when (it) {
+                    CLOSE_WITHOUT_SAVING -> {
+                        runOnUiThread {
```
tibbi, line 254: why are you creating a new thread above, then switching back to UI thread without doing anything? Create a new thread only at saving an event

tibbi:
is this dialog really always shown? Even without checking, if any field has been filled in?

fnrngg:
Sorry for the first one, it's my first PR and didn't notice such things. For the second, at first I wanted to show only if some changes were made, but then I checked it on my mobile's calendar and it anyway asks "Discard changes?", so I left without checking. Fixed both of them in the last commit.

tibbi:
I still get the Save dialog always, even without changing anything

fnrngg:
On my device it always shows dialog if creating new event, but if editing existing one shows only if something changed

tibbi:
it shouldnt be shown at creating a new event either. It is pretty bad user experience

tibbi:
also take the Discard dialog from Simple Draw, no need to invent a new dialog, or anything else. Just reuse the existing solutions wherever possible.

fnrngg:
Fixed both, but I don't know if it's bug or not, but when I call showEditRepeatingEventDialog() while saving I can press back or tap on screen and stay on edit scene, but mEvent stays changed and if I press back button again without changing anything, it can't show "discard changes". So is it bug or I have to save copy of starting mEvent?

tibbi:
Im not sure what you mean, I will have to take a look at it. But it sounds like a specific use case already, shouldnt happen often.

tibbi:
looks good like this, hope it will work well :) Thanks

PR #1155, opened 2020-06-06
galician (gl) update

xmgz:
fixed some errors, updated new text chains and improved concordance


tibbi:
the short app Google Play description can have max 80 characters, yours has 84 even with &amp; replaced with &. Make it shroter please, Im using the old short version for now.

Issue #1080, opened 2020-02-13
[Feature Request] Adding a divider between months in simple event view


v1r0x:
I switched from google calendar to your app and I really love it!
But I miss one small feature which really helps me when looking for events. I google calendar is a divider (very large) between two months. Something similar (and not as large) would be a nice feature :)

tibbi:
some UI changes will be added in 6.16.0

PR #1171, opened 2020-07-06
Adding a divider between months in simple event view #1080

Honk2:
add a divider with accent color
Not exaactly #1080 but just a different color and more subtile than a larger gap


tibbi:
Hey, I dont really like it, it looks weird. That view needs some bigger UI rewrite in general. Thanks though

PR #1180, opened 2020-07-30
Added Basque translation

avtkal:


tibbi:
create only the res/values-eu/strings.xml file please, do not add anything else. Maybe create a new pull request, will be easier than deleting it

avtkal:
> create only the res/values-eu/strings.xml file please, do not add anything else. Maybe create a new pull request, will be easier than deleting it

I think I changed it correctly. Please let me know if I should do anything else.

tibbi:
yeah looks good, thanks :)

PR #1181, opened 2020-07-31
Fixes #1170 Contact birthday not correct on changing timezone

### LINKED ISSUE #1170: Contact Birthday not correct when changing timezone
Good Day, 
when I am switching timezone I noticed that birthdays are shown on the "wrong" day.
I understand why that it is happening but I feel it does not make sense and can be confusing.
The current workaround I applied is to "add contact's birthdays" again when I am in a different timezone.

Best regards

Varsha-29:
The birthday was captured in DB with current time zone. While retrieving the date was parsed based on the current time zone. Modified that logic to parse date based on the timezone at which it was originally captured


tibbi:
was the bug occuring only at the monthly view, or everywhere? I cannot really reproduce it

tibbi:
woops, didnt mean to accept it yet

PR #1185, opened 2020-08-14
Display Location if description is empty

### LINKED ISSUE #1177: Show location in widget if description is not set
Currently in the widget each calendar entry looks like this:
```
08:00   Title
19:00
```

In the bottom line it shows the description when there is one, however for the entries I sync via a URL (university time schedule) there is no description, there is however a location associated with each entry.

So I would suggest showing the location where the description would be if the description is empty:
```
08:00   Title
19:00   Location: Room 123
```

The `Location` prefix would just be the text of the current location input placeholder in the edit dialog, which is already localized.

xerus2000:
Haven't tested it in the actual app, but seems simple enough

Fixes #1177


tibbi:
I dont think this is the expected behaviour, to just automatically replace the description with location. It might need another setting or so, we will see. Thanks though

xerus2000:
@tibbi you should be careful with adding new settings. Can you name a usecase for not wanting this behavior?

tibbi:
I want to avoid adding such smart solutions. If I use event descriptions and Im used to seeing the event description under the event, I dont want to have it mixed with the location.

PR #1186, opened 2020-08-14
support private event classifiers

### LINKED ISSUE #1022: Unable to set event as private
Is this feature already there and did I miss it somehow ?
I don't see it neither while editing any event nor when scanning the main app's configuration...
(Btw, great job on this happ, most thankfull !)

romor:
This fixes #1022.


tibbi:
can you mark an event as Private in the Google Calendar app? I cant see it. I still cant decide if I want that extra switch there, people arent really requesting this at all.

romor:
> can you mark an event as Private in the Google Calendar app?

Yes, you can. 
I think I cannot attach the screen-shot here to this pull-request. Just tell me if you like to get it via email.

tibbi:
I see the private events on the website, but not in the app

romor:
> I see the private events on the website, but not in the app

There are also available in the calendar app. Just sent you the screen-shots via email.

tibbi:
I see, so it is apparently visible only at those calendars, that are shared with someone. Thats something you could research and add too, so the lock wont be visible at not shared calendars. And it definitely shouldnt be visible at local events that arent even CalDAV synced. I just want to avoid showing extra elements, they flood the UI. Thanks

tibbi:
closing it until implemented properly

PR #1204, opened 2020-09-25
Added missing italian translations for the app

MhzDev:
Translated some strings in Italian that were still in English


tibbi:
nice, thanks

Issue #1081, opened 2020-02-29
[bug] Februar 29th


SimonHaas:
Hi, contacts (created through your simple contacts app) with a birthday at Februar 29th are displayed on  Februar 28th.

VinamraGuptaa:
Hello. I would to work on the issue. Is there any Pre requisites or installation guide?

SimonHaas:
Hello, it is kind that you want to tackel this issue. 
In order to reproduce the bug it should be enough to create a contact with the Simple-Contacts app, or any other contacts app, with a birthdate of february 29th 1988 for example. In the Simple-Calendar app this birthday currently shows on february 28th 2020, provided you activated "Add Contact birthdays" via the options menu in the app.
Regarding the installation for development of the app it should be as easy as cloning this github-repository than opening and starting it in Android Studio.
If you have any trouble setting up the project you should be able to get it up and running with the help of an general guide on how to develop Kotlin apps. If not, then just ask here.
Best regards and good luck ;)

Issue #1056, opened 2020-01-06
Check a leap year related issue


tibbi:
more info at email from Victor on 6th January

Honk2:
Hi Tibbi,

could you please share the issue description? Is it some how related to #1081 ?

PR #1207, opened 2020-09-27
Februar 29th #1081 and Check a leap year related issue #1056

Honk2:
This patch will handle the 29th of February correctly on yearly recurring events.
On non leap years the event will be on the 28th but on the 29th on leap years.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/models/Event.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/models/Event.kt
@@ -46,7 +46,7 @@ data class Event(
 
     fun addIntervalTime(original: Event) {
         val oldStart = Formatter.getDateTimeFromTS(startTS)
-        val newStart: DateTime
+        var newStart: DateTime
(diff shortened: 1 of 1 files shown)


tibbi:
lets rather hide such events on nonleap years and only show them if february 29 is available. Thats how it works in Google Calendar and thus people expect that behaviour.

Honk2:
I just realized, that I did things wrong.
On non leap years, a recurring event originally on Feb 29th, must move to March 1st.
I do not know the google calendar, but think that it works like that. Therefore I need to modify the patch....

Honk2:
ok, now the behaviour is as it should be.
At least Birthdays are now shown always on the correct day.

tibbi:
I dont want to show Feb 29 events on March 1, it just sounds bad. Lets stick to the way Google Calendar works in such cases, people are used to that.

Honk2:
ok, I respect this. Although it seems that in most countries, the 1st of March is the legal data for time spans in non leap years.
Funny that the google calendar behaves like that.

PR #1232, opened 2020-11-13
Update mexico.ics

juacq97:
Hi! I've updated mexico.ics since it's going to expire this year. I extended it until 2025. Besides that, I changed some dates and holidays names. In Mexico we have two kind of holidays: the actual holiday and the derived day off. For example, next november 20th is a holiday, the Revolution's Day, but it's a working day, the day off will be the monday 16th to match the weekend and have a "long weekend", weird I know. The past ics file used the days off instead of the actual holiday, I changed it because it will be strange if my calendar reminds me about a holiday when there's no actual holiday, or uninformative for any non-mexican who would like to know our holidays, and we all know when we have a "long weekend".
I added a buch of new holidays too, not just thew one that have days off, like Mother's day and Flag day. These are working days, but still holidays. I followed the same format of the last ics file and I imported it to the app without problems.


tibbi:
Hey, have you tried also if they import well and show well?

juacq97:
Yep, it shows well and there's no problem when importing it. 

image
image

tibbi:
ok cool, thanks

PR #1259, opened 2021-01-09
Added during day reminder for all-day events

correia55:
I added support for the user to create a reminder for a specific time, during the day of an all-day event.
Also added this option to birthdays and anniversaries.


correia55:
Further testing of the changes showed some problems, which seemed to be related with the SQL query.
After the correction, reminders worked for both birthdays and normal all day events.

The event menu will show up exactly the same as before, when the all day toggle is off.
When it is toggled on, a new section appears.
Added screenshots of the changes.

all_day
all_day_reminder
birthdays

tibbi:
the idea is good, but the implementation has to be changed. The "Add during day reminder" field has to be gone, it has to use the standard 3 places for reminders, nothing extra. It could be stored at the reminder type, without creating a new db column. Reminders have "before" at them, that could be replaced to "On the day at 9:00" or so.

correia55:
Just to make sure I'm understanding it correctly, what you want is:
- when clicking on one of the existing reminders, we would have an extra option, only for all-day events,  "During day", which, similarly to the "Custom" option, would open a second window with the time selector window;
- we would have a new reminder type, which would then allow for the values of minutes to be negative, in the same way as they are in my implementation. This also includes the SQL query that I changed in my previous commit.

tibbi:
1. Im not sure about the UX at selecting a reminder yet, but ye, I guess above "Custom" there could be a button like "During day" or so, then on click a time picker shows up. The label will then be "During day at 9:00" or so.
2. if negative numbers arent used for anything else, then we dont even need to create a new reminder type, just use negatives. I wasnt sure in negatives now, havent worked with that code for a long time. There could be some issue with -1 as "no reminder", but I guess we can keep "-1" as no reminder. So it will be impossible to set a reminder during day for 00:01. Should be good enough.

correia55:
I think I have completed the implementation, I'll run some more tests in the following days just to make sure everything is alright.

Nonetheless, there's some points I would like to discuss:
- most of the changes now happen in the Simple-Commons (as that is where the seconds picker dialog is), with only some minor changes to verifications being done on the Simple-Calendar;
- the function getChoppedList is still being used by the Simple-Calendar, even though it was removed in the latest version of the Commons. I copied the latest version of this function to the Simple-Calendar just so everything could work normally, but what should I do for the pull request?

tibbi:
you can create pull requests in Simple Commons too. getChoppedList has been replaced by a built in function like .chunk or so, try it and see what autocompletes you get

correia55:
The part in the Simple-Commons is in: https://github.com/SimpleMobileTools/Simple-Commons/pull/1027

Adding "during the day" reminders for events on the day they are added are working for both normal events, as well as anniversaries and birthdays. Nonetheless, these don't seem to work when the events are in a different day.

I've commited both the changes to Simple-Calendar and Simple-Commons.
I think the commit in Simple-Commons is correct.
I've run through the entire thing a few times, maybe someone else could figure out what it is that is wrong.

tibbi:
Lets allow setting reminders to minutes only, not seconds. I declined the seconds picker, it is way too advanced and unnecessary.

correia55:
I am a bit confused, I did not add a new function, the PickSecondsDialog was already the dialog being used for choosing the reminder. The only thing I did was add a new option to it.
Given that the new option is picked from a TimePickerDialog, the user will only be able to pick a reminder in terms of minutes. The only reason I'm using the functions that work with seconds was to minimize changes.
If this is not what you want, what should I do?

correia55:
For the text, it does call getFormattedMinutes, which then calls getFormattedSeconds. In this case, I could indeed add an option to verify if the minutes are negative and, in this case, get the text directly, without calling getFormattedSeconds. Nonetheless, I think that this is a less elegant solution. Let me know what you think.

tibbi:
right, I was just reading the code quickly, thought you added seconds. Ill take a look at the code itself and try it soon

tibbi:
have you added only the UI part, but not the actual functionality with the reminder being set up?

tibbi:
ah damn I accepted it unintentionally, guess Ill have to finish it now

correia55:
What is there to finish?

correia55:
Even though I changed the version of the dependency for the Simple-Commons, it does not seem to have been changed.
Is that what you meant by finishing?

tibbi:
no, tweaking the actual functionality. Like if you have set the during day reminder, avoid adding as the default one for the next event which isnt even all-day yet. Or adding a generic check to really avoid saving such reminders to non-all day events. And maybe others, I havent really tested it yet.

correia55:
Ok, I can try to improve it tomorrow.

tibbi:
it is already done, I added some tweaks. Thanks :)

PR #1288, opened 2021-02-24
Add option to only list the start time of events in the event list widget

ClearlyClaire:
Currently, upcoming events that aren't all-day events take at least two lines in the event list widget.
In an effort to make the widget more compact, this PR adds an option to only list the start date/time of each event.

## Before / After

image
image

## Configuration screen

image


tibbi:
Ill close both pull requests until added properly in the Widget Config screen, more things should be moved there to avoid flooding the settings screen.

PR #1289, opened 2021-02-24
Add compact widget mode

ClearlyClaire:
Add a setting to make the more compact by removing the header as well as some padding.

## Before / After

image
image

And, on top of #1288:
image


tibbi:
move both of the requests into the Widget configuration screen please, they affect only one widget, so no need to fill the settings even more

ClearlyClaire:
This seems significantly more involved, I will have a look later.

tibbi:
Ill close both pull requests until added properly in the Widget Config screen, more things should be moved there to avoid flooding the settings screen.

PR #1293, opened 2021-02-28
Update strings.xml

smarquespt:


app/src/main/res/values-pt/strings.xml
```diff
@@ -189,7 +189,7 @@
     <string name="last_view">Última utilizada</string>
     <string name="new_events">Novo evento</string>
     <string name="default_start_time">Hora inicial padrão</string>
-    <string name="next_full_hour">Próxima hora</string>
+    <string name="next_full_hour">1 hora</string>
```
tibbi, line 192: I dont think this is correct. It is supposed to say "The next full hour", not "In 1 hour". Any thoughts?

smarquespt: Yes, you're correct. Merge it and I roll back after.

Issue #1296, opened 2021-03-06
[Bug] Incorrect Indian holidays


mohitsoni-dev:
Many Indian holidays for year 2021 are wrong in the calendar. Examples:
Actual date for Holi is 28 March IST in the app it is 23
![](https://user-images.githubusercontent.com/59525097/110197045-cdd93a80-7e6e-11eb-8b5d-cd8a1cfe79f5.png)
Mahavir Jayanti is on 25 Apr and Sri Ram navami is on 21 Apr but these events are on wrong dates in the app.
![](https://user-images.githubusercontent.com/59525097/110197087-0f69e580-7e6f-11eb-8e81-295715c34815.png)
There are many more wrong holidays.

tibbi:
The events are stored in https://github.com/SimpleMobileTools/Simple-Calendar/blob/master/app/src/main/assets/india.ics , can you fix it there? That would be the simplest. It should be quite straightforward how do the events work there, they are separated by "BEGIN:VEVENT" and "END:VEVENT". You can open the file with any text editor, but you can edit it on the website too. The instructions can be found at the guide for translating is at https://github.com/SimpleMobileTools/General-Discussion#how-can-i-suggest-an-edit-to-a-file , but use the previously linked file. Thanks

mohitsoni-dev:
Sure I will fix it. Thanks.

PVTejas:
This is a problem every year, because Indian holidays follow a lunar calendar. Can we have the Indian holidays calendar not set to "repeat yearly"?

PR #1297, opened 2021-03-06
For #1296 - Corrected major Indian holidays dates

mohitsoni-dev:
As Indian holiday follow hindu calendar, india.ics file needs to be updated yearly. I have corrected most of the major dates that I could. Thanks.


app/src/main/assets/india.ics
```diff
@@ -215,60 +215,60 @@ END:VEVENT
...
 RRULE:FREQ=YEARLY;INTERVAL=1
 EXDATE:20170826}
 END:VEVENT
 BEGIN:VEVENT
 SUMMARY:Solar Eclipse [Annular]
 UID:ind_31
-DTSTART:20160901T070800Z
-DTEND:20160901T071200Z
+DTSTART;VALUE=DATE:20160901
+DTEND;VALUE=DATE:20160901
```
tibbi, line 255: should end with 0902 no? If it lasts only the first day, just like other events

mohitsoni-dev: Solar Eclipse last for a couple of hours only that's why. Anyway, I will try to find the exact time of the eclipse and fix it up. Thanks.

tibbi: right, should be good like that. If its just a few hours, it should have DTSTART and DTEND labels too, for start and end time. You can see some example if you try sharing an event. Anyway, this looks ok, thanks

PR #1299, opened 2021-03-07
update Japanese holiday

ikazuhiro:
Please update Japanese holiday information.

- I keep obsolete holiday by adding `UNTIL` rule in `RRULE` property, beacause I don't know how old events you should keep.
- Holiday's Japanese name of [jap_56](https://github.com/ikazuhiro/Simple-Calendar/blob/79f9b5d158a3b4c03615bdd2ad35d0d9dcf5a37e/app/src/main/assets/japan.ics#L414) and [jap_57](https://github.com/ikazuhiro/Simple-Calendar/blob/79f9b5d158a3b4c03615bdd2ad35d0d9dcf5a37e/app/src/main/assets/japan.ics#L421) may be are not correct.  I copied from other [calendar project](https://github.com/yoshinari-nomura/mhc/blob/master/samples/japanese-holidays.mhcc), but Japanese government's document simply say 'holiday'.


tibbi:
alright, should be good :) Thanks

PR #1300, opened 2021-03-09
fix #1201, support 5th weekday recurring events

### LINKED ISSUE #1201: [Bug] Scheduling recurring CalDAV events
I use calendar extensively and recently switched from DigiCal (an excellent calendar manager) to SimpleCalendar. Yesterday I started to notice that events were not correctly being synced to SimpleCalendar over CalDAV. I use DAVx5 btw.

Here is the exact problem:

- On my computer, I use Evolution to create an event that occurs every 5th Wednesday starting on September 30th, 2020. I set this event to only repeat twice.
- On Evolution, I also check that event is created once for September 30th, 2020, and the next event is created for December 30th, 2020. I manually verify that these are the only days and within the range I specified which have a 5th Wednesday.
- On my phone when I refresh the event in SimpleCalendar, it created the one event for September 30th, 2020, **but it also creates another event for October 28th, 2020!**
- On another phone which also has DAVx but using DigiCal, I check my Calendar and the event is created as on Evolution: September, and December

I am using the latest version of Simple-Calendar Pro v6.10.0

**EDIT:**

I should also add that deleting the event on Evolution, also deletes the correct events on SimpleCalendar, so the problem is definitely with the way it displays recurring events.

Upon further testing my conclusion is that instead of showing the event every 5th Wednesday, it defaults to showing it as every last Wednesday, which is not the same thing

**EDIT2:**
Looks related to #963

ikazuhiro:
This fixes #1201 and enables you to input such events on application.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/EventActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/EventActivity.kt
@@ -644,17 +644,9 @@ class EventActivity : SimpleActivity() {
     private fun getAvailableMonthlyRepetitionRules(): ArrayList<RadioItem> {
         val items = arrayListOf(RadioItem(REPEAT_SAME_DAY, getString(R.string.repeat_on_the_same_day_monthly)))
 
-        // split Every Last Sunday and Every Fourth Sunday of the month, if the month has 4 sundays
+        items.add(RadioItem(REPEAT_ORDER_WEEKDAY, getRepeatXthDayString(true, REPEAT_ORDER_WEEKDAY)))
         if (isLastWeekDayOfMonth()) {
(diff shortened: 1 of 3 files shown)


tibbi:
what is the use of this? There is already an "every last Monday" etc option

ikazuhiro:
I want to schedule every 5th weekday's event, which skips a month which doesn't have 5th weekday.

For example,
```
Every 5th wednesday
 -> 2021/03/31, 2021/06/30, 2021/09/29, 2021/12/29, 2022/03/30...
```
```
Every last wednesday
 -> 2021/03/31, 2021/04/28, 2021/05/26, 2021/06/30, 2021/07/28...
```

I can make schedules by excluding "every 4th wednesday" from "every last wednesday" manually, but It is very annoying.

tibbi:
okay looks good from my tests, hope it wiill work well. Thanks

PR #1325, opened 2021-04-02
fix(au-holiday): misc fixes 

curbengh:
- deduplicate recurring holidays
- add 2022 holidays
- remove 2020 (non-recurring) holidays

https://info.australia.gov.au/about-australia/special-dates-and-events/public-holidays


tibbi:
cool, thanks

PR #1333, opened 2021-04-05
fix #1332 : concurrent zero duration in week view

### LINKED ISSUE #1332: concurrent zero duration event not shown correctly in week view
If there are zero duration events overlapping with other events, these will still be draw over the full column width.
To reproduce:
Set up an event A  e.g from 2pm to 4pm and event B from 3pn to 3pm on the same day.
event A will be half column with, but event B full column width.

Honk2:
The range comparison did not include the equal case.
Changed < to <= and > to >=


Honk2:
I found an issue with this pr
Adjacent events are handled as concurrent..

PR #1339, opened 2021-04-10
fix #1332 concurrent zero duration in week view

### LINKED ISSUE #1332: concurrent zero duration event not shown correctly in week view

Honk2:
Instead of changing the range collision, it is better to give zero duration events a minimum duration.


tibbi:
what does this fix?

Honk2:
It gives zero duration events (Start time equals end time) a minimum duration of 1 minute.
Because the range event collision check fails if the start and end time are the same.
see Range.kt 
> fun Range<Int>.touch(other: Range<Int>) = (upper > other.lower && lower < other.upper) || (other.upper > lower && other.lower < upper)

Changing the range check to include equal like:
> fun Range<Int>.touch(other: Range<Int>) = (upper >= other.lower && lower =< other.upper) || (other.upper >= lower && other.lower =< upper)
is not good, because if the end time of an event is the same as the following start time, it will be treated as concurrent.
As events have a minimum display height, this should do.

I have not checked if an unbalanced touch would do.

Issue #1241, opened 2020-11-25
Events over multiple days


sinus-x:
Hi.
When there's an event that ends the next day, it is drawn on the top, with other events that are set as *All-day*.

I keep finding this a bit confusing (I expect them to be at the time they are supposed to be at). Would it be much of a problem if they were drawn as any other?

(Illustrated here by the blue event)
![calendar](https://user-images.githubusercontent.com/14353790/100248618-fc747a80-2f3b-11eb-9382-44c4390dde2e.png)

dumblob:
I experience the very same confusion. The suggested change would be perfect. Thanks @sinus-x !

dumblob:
Oh, I misread. I was expressing myself to all-day events which I would prefer to see the same way as on the right picture - i.e. **the very same** as non-all-day events which though span multiple days.

Shall I make a separate issue for this?

tibbi:
not sure what you mean here, you can already make it so that all-day events or ones spanning across midnight wont be at the top separate bar

dumblob:
Ok, then it is a bug in *6.18.2 Pro* because if I have unticked "Show events spanning across midnight at the top bar" checkbox in Settings, then I can still see all-day events in the top bar (but non-all-day events spanning across midnight are correctly shown as in the right picture above).

I will add a screenshot in a minute to demonstrate it.

PR #1377, opened 2021-05-15
implements Events over multiple days #1241

Honk2:
Show events spanning midnight in weekview at top (default) or vertical from start time to

This time clean and lean


tibbi:
looks good, shouldnt break anything :) Thanks

PR #1386, opened 2021-06-02
Fix: popup window closes on orientation change

vbh:
Steps to Reproduce;
1. Click on Action Menu items button
2. Rotate the screen
3. Popup window is closed


tibbi:
I dont want to mess with that, it can affect quite a few other things too

Issue #1376, opened 2021-05-14
FR: Disable warning about notifications not enabled


JohnVeness:
I use https://f-droid.org/packages/com.github.quarck.calnotify/ to handle calendar notifications, rather than whatever calendar app I use (I switch between Etar and Simple Calendar). In Simple Calendar, whenever I add or change a notification in a calendar entry, I get a warning "Notifications of this application are disables. Please go in your device settings for enable them".

Could you add an option to never show this warning? Either in the app's settings, or a "Never show this again" tickbox on the warning maybe?

tibbi:
Any steps for reproducing this? I see no warnings from the Calendar Notifications app. It definitely doesnt come from us.

JohnVeness:
Android settings:
![](https://i.imgur.com/njHECDA.png)

App after clicking the reminder line in a calendar entry:
![](https://i.imgur.com/g66J9Ob.png)

App after clicking OK:
![](https://i.imgur.com/4jojNeF.png)

To be clear, other than the warning message, everything works fine - I can set the reminders and they reminders show up correctly (from my other app that is handling them). I can understand why you want to show the warning, as I guess most users would want to know, but certainly for me I'd love if there was a "Don't show again" tickbox.

tibbi:
oh right, I see. I even forgot about that message. Ye, I will try adding the tickbox.

PR #1392, opened 2021-06-10
#1376, Add setting to disable warning about notifications not enabled

tanvir-ahmod:


tibbi:
why would something like that be needed... It just shows a dialog for the first time.

tanvir-ahmod:
Actually the dialog shows every time if you want to change the reminder and your notification is disabled.  Perhaps thats why this feature  was requested as I can see in issue list.

Issue #1411, opened 2021-07-17
Quick calendar filtering


chaptergy:
If you have calendars from e.g. family members or coworkers, you might want to usually hide them and only periodically show them. For now you have to go though the overflow menu and filter the events by type, which works but could be more efficient. For example with a custom row of toggleable calendars at the bottom. This should be optional and probably disabled by default.

Basically the same thing as the second paragraph from https://github.com/SimpleMobileTools/Simple-Calendar/issues/1062.

This issue is less a feature request for you to implement but instead I'm asking whether you are open to the idea, and would accept a merge request if I get around to adding this.

tibbi:
well if it gets implemented nicely and disabled by default, I can add it

PR #1412, opened 2021-07-17
Adds quick filters for event types to the bottom of the screen

chaptergy:
This PR adds an optional quick filter of event types / calendars to the bottom of the screen. You can quickly enable and disable a calendar there. You have to select which calendars should show up in this quick filter bar, and it will not show up if there are none selected. By default there will be none selected, and new calendars will not automatically be added. It will always take the entire width of the screen, and if it becomes too wide, will become scrollable horizontally.
If the name of a calendar becomes too long, it will be ellipsed after two lines.

image

Resolves https://github.com/SimpleMobileTools/Simple-Calendar/issues/1411.


tibbi:
ye looks good in general, Ill tweak a few things here and there. Thanks

Issue #61, opened 2016-12-09
Add birthdays/anniversaries with calculating age/years


tibbi:

Docbroke:
Also add tags, for holiday/birthday/anniversary etc., with automatice yearly recurrence for birthdays/anniversaries

Anke:
+1
This feature request dates back to 2016!? I'd also love to see a persons age behind a birthday.

duesseldorferjung:
+1

tibbi:
will be added in 6.15.1

PR #1442, opened 2021-08-08
Append age/years for birthdays/anniversaries to event title

KryptKode:
**Notes**
- this PR appends the age/year for birthdays/anniversaries to the event title.
- it addresses this [issue](https://github.com/SimpleMobileTools/Simple-Calendar/issues/61)
- we need to account for when the birthday/anniversary year is not specified in the contact data so we do not show an incorrect age/anniversary. Still trying to reason around this hence I am submitting a draft pull request. 
Screenshot

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Constants.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Constants.kt
@@ -93,6 +93,7 @@ const val REPEAT_ORDER_WEEKDAY = 4                      // i.e. every 4th sunday
 // special event flags
 const val FLAG_ALL_DAY = 1
 const val FLAG_IS_PAST_EVENT = 2
+const val FLAG_MISSING_YEAR_EVENT = 4
 
 // constants related to ICS file exporting / importing
 const val BEGIN_CALENDAR = "BEGIN:VCALENDAR"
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Constants.kt
```diff
@@ -174,3 +175,13 @@ fun isWeekend(i: Int, isSundayFirst: Boolean): Boolean {
         i == 5 || i == 6 || i == 12 || i == 13
     }
 }
+
+fun getDateFormatsWithYear() = arrayListOf(
```
tibbi, line 179: where did you get these formats from? You could reuse the ones from https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/kotlin/com/simplemobiletools/commons/helpers/Constants.kt#L286 , I haven't seen other ones so far. Or maybe add the extra ones in Simple Commons too, if it makes sense.

KryptKode: Actually got them from [there](https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/kotlin/com/simplemobiletools/commons/helpers/Constants.kt#L286 ). I would add them

Issue #1260, opened 2021-01-12
Change calendar color


kristof-linux:
Hello, I have DAVx5 and nextcloud hosting and like to have all my calendars the same color. It seems Nextcloud is not supporting calendar colors over CalDAV. If i set in DAVx5 disable 'manage calendar colors' and try to change the color with Calendar Color F droid, the color changes in my google calendar app, but not in the simple-calendar app, it stays with the "DAVx5 green". Is it possible to solve this? Thanks a lot!

tibbi:
what Simple Calendar app version are you using?

kajdo:
Same here - just tried it out with the same tools as OP (nextcloud - sync via DAVx5 (green) - change tried with Calendar Color F-droid - nothing changed)

the version I'm using is v6.13.6 Pro

kajdo:
just a tip (figured out how to solve it for me and might be useful)

in DAVx5 - I configured (sorry - german ... but you'll figure it out)
- Kalenderfarben verwalten YES
- Unterstützung für Terminfarben YES

plus -- and this did the trick for me ... I did "Kalender neu erkennen" (3 dots-menu in the main navigation in the CALDAV tab)

I resynced my calendar  -- and it worked ;)

PR #1486, opened 2021-09-30
fix updating event type colors for caldav events

KryptKode:
**Notes**
- fix updating event type colors for caldav events 
- add `color` field to `Event` class so it's included in the computation for hashCode, hashCode is used to determine when to update events
- should address this [issue](https://github.com/SimpleMobileTools/Simple-Calendar/issues/1260) 

https://user-images.githubusercontent.com/25648077/135476903-561841b2-0db6-4382-9a6b-c7097396fd30.mp4


tibbi:
Hm? Does that update the colors on the server side?

KryptKode:
It does, but Google Calendar resets the color back to what it had stored

tibbi:
Im not sure, thats what needs research. The last time I checked it worked just fine at syncing via DAVx5 or so, it is just Google Calendar having issues.

KryptKode:
Please check @tibbi . 
Now it would update the color on the other Calendar apps when they are refreshed

tibbi:
it actually works well to me, we will see the feedback. For now I will remove that top item at the color picker which allow updating the color locally only. Thanks!

PR #1491, opened 2021-10-06
Start the weekly view with the current date

### LINKED ISSUE #1212: Feature request: improve the scaled weekly view
I really like the idea of having "1 day" view as a list and a scaleable "Weekly" view: given that I live a busy life and need to constantly resolve clashes in my calendars a view that shows overlapping events is a necessity.  So, I tried to achieve this by switching to the "Weekly" view, but scale it down to just a single day.  I looks great...

However, there is a usability problem.  My week starts on Sunday (it's a choice), so when I tap on the "Today" button at the top, the "Weekly" view switches to the week where "today" is, but since I have just one day of the week in my viewport it is always Sunday's schedule I see.

The idea, I think, that is worth implementing is that if the scale of the "Weekly" view is less than 7 days, show the days in the view starting from 7 minus the current scaling factor, e.g. if I have the scaling factor of "3 days" and "today" is, say, "Wednesday" (with week starting on "Sunday"), then when today is tapped it should show 7 - 3 => days 4 (Wed), 5 (Thu), and 6 (Friday).  This assumes that day 1 is Sunday.

I think it would be a really wonderful feature to have and would make the "Weekly" view the most used out of other views.

0x8664b2:
Fixes #1212 

I second the proposals in #1212 however I also see an argument for starting the weekly view with the first day of the week, but only when the sliding scale is set to a full 7-day week. 

I can rework this with your guidance if you rather want to move this into a different direction


tibbi:
or guess Ill do that myself, thanks

Issue #906, opened 2019-06-04
Auto refresh to add Phone contacts Birthday and Anniversary


s0uthk1ng:
I do not see any option to auto sync my phone contact's Birthday and Anniversary. Every time i update or add a contacts birthday i have to manual click on "Add contact birthdays/anniversaries".

thenktor:
Just wanted to add the same feature request.

tibbi:
will be added in 6.16.0

PR #1527, opened 2021-10-29
Added auto sync for birthdays and anniversaries on the app start (#906)

Aga-C:
Hi,

I've added an option for automatic sync of birthdays and anniversaries on the app start. It's optional, and there's a permission check while toggling an option. If user revokes the permission, it doesn't throw any error, just ignores adding new events.

https://user-images.githubusercontent.com/85929121/139481278-fd0bc1f7-28d3-44be-a9e7-ac1179be9f88.mp4


app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/MainActivity.kt
```diff
@@ -589,6 +593,37 @@ class MainActivity : SimpleActivity(), RefreshRecyclerViewListener {
         }
     }
 
+    private fun addBirthdaysAnniversaries() {
+        if (!hasPermission(PERMISSION_READ_CONTACTS)) {
+            return
+        }
+
+        val reminders = arrayListOf(REMINDER_OFF, REMINDER_OFF, REMINDER_OFF)
```
tibbi, line 601: the user can already select reminder at importing birthdays and anniversaries, lets reuse the values he selected here too

app/src/main/res/values-ar/strings.xml
```diff
@@ -202,6 +202,7 @@
     <string name="highlight_weekends">ظلل نهايات الأسبوع في بعض العروض</string>
     <string name="allow_changing_time_zones">Allow changing event time zones</string>
     <string name="manage_quick_filter_event_types">Manage quick filter event types</string>
+    <string name="add_bdays_anniv_automatically">Add birthdays and anniversaries automatically after launching the app</string>
```
tibbi, line 205: let's change this to "Add new birthdays automatically" and "Add new anniversaries automatically"

app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/MainActivity.kt
tibbi: People will expect having the last used reminders there. Just store the 3 reminders in some new shared pref config field.

app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/SettingsActivity.kt
```diff
@@ -621,6 +623,34 @@ class SettingsActivity : SimpleActivity() {
         }
     }
 
+    private fun setupAddBirthdaysAutomatically() {
```
tibbi, line 626: oh, this whole thing shouldnt be a settings item, just a checkbox in the importing dialog

Aga-C: Ok, so just to confirm:
- user taps on "Add contact birthdays" from menu
- dialog with setting reminders is displayed, and despite reminders user also has a checkbox "Add new birthdays automatically"
- if checkbox is selected, we also store the reminders setting in config
- rest works the same as now - user starts an app, and we check for new birthdays
Of course, the same logic would be applied to anniversaries too.
Am I right?

tibbi: yep, like that

app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/MainActivity.kt
```diff
@@ -589,6 +591,41 @@ class MainActivity : SimpleActivity(), RefreshRecyclerViewListener {
         }
     }
 
+    private fun addBirthdaysAnniversariesAtStart() {
+        if (!hasPermission(PERMISSION_READ_CONTACTS)) {
```
tibbi, line 595: you should check config.addAnniversariesAutomatically and config.addBirthdaysAutomatically first, those are a lot quicker and simpler checks than some permissions

Aga-C:
I've applied changes.

image

tibbi:
oh gosh, stop showing a toast message at every app launch, it is annoying

Aga-C:
It shows toast only when there is a new anniversary or birthday in contacts. I've tested it again now, and it doesn't show every time.

tibbi:
It was actually showing everytime to me with a success message. Guess it is some unrelated issue then, it doesnt recognize existing events or what. Will take a look at it again.

tibbi:
okay I removed a corrupt contact, should be good I guess. Thanks

PR #1549, opened 2021-11-22
Spelling corrected

10cents:
I  have corrected a spelling, but I do not understand the difference between reminders and notifications, or how I can choose between them? Could you help me here please - just to be sure to make a meaningful translation.


tibbi:
Well reminders are usually used by calendars or so to actually remind you something. Notifications are used for notifying you, for example if you receive an SMS. But ye, visually it is the same thing and in some languages it might be called the same, no big deal.

PR #1561, opened 2021-11-28
Verb or noun?

10cents:
Verb or noun? Here I have translated 'Filter' as a noun, but it could be a verb, what is correct? Many English words can be used as both nouns and verbs, so the translation can be hard outside the context.


tibbi:
it actually refers to the above string filter_events_by_type "Filtrer begivenheder efter type", but I didnt want to paste the whole thing as this new string will be a toast shown for a couple seconds only and it is already too long

PR #1567, opened 2021-12-04
Update strings.xml German language version

sphenotux:
Wording changed in line 160; translated from English into German in line 172.

DISCUSSION needed on line 167, now saying:
"    <string name="day_end_before_start">Terminbeginn kann nicht vor Terminende liegen</string>" - from my point of view translation is the other way round: Terminende kann nicht vor Terminbeginn liegen.

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -157,7 +157,7 @@
     <string name="manage_event_types">Termintypen verwalten</string>
     <string name="start_day_at">Wochenansicht beginnt um</string>
     <string name="end_day_at">Wochenansicht endet um</string>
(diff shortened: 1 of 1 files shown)


Honk2:
You are right, the current string does not make sense.
The correct translation is: "Terminende kann nicht vor Terminbeginn liegen."
Although I would prefer "Terminanfang" over "Terminbeginn"

tibbi:
so should I accept this PR or not?

tibbi:
nevermind, there is now a conflict because of the Weblate update, create a new PR if something needs to be changed

sphenotux:
Thanks, will go through this again - probably at the weekend, and get back to you.

sphenotux:
I see this has been fixed in 5ba475de0feaf2479b5f7768c0ba7e3b80fddae1 on December 8 by @Honk2 - thx. Will delete my local branch.

PR #1577, opened 2021-12-12
Fix: Wrong days highlighted in week view

B14CK313:
If `startWeekWithCurrentDay` was enabled, `highlightWeekends` lead to the wrong days beeing
highlighted, because the column index passed to `isWeekend` didn't correspond to the day of week
anymore. This change fixes this by using `curDay.dayOfWeek` instead.


Aga-C:
Will it highlight Saturday and Sunday, when it's set to start the week from Sunday? I see that you've negated the config value for this, that's why I'm asking.

B14CK313:
You are right, I incorrectly assumed that if the week starts on Sunday, the weekend is shifted as well.
Fixed it by just passing `true`, because the return value of `curDay.dayOfWeek` uses 1 to 7 for Monday to Sunday (https://www.joda.org/joda-time/apidocs/org/joda/time/DateTimeConstants.html#MONDAY) instead of starting to count at 0 like this code does. Thus, passing `true` makes it handle 1 as Monday.

PR #1582, opened 2021-12-18
Added keeping date when changing views (#1403)

### LINKED ISSUE #1403: Keep selected date when changing view
Hi Tibor,

not sure if duplicate or denied already by you, can't find an open issue at least. 

E.g navigating in monthly view to a future month, tap on a day to see it's details, and switch then to weekly view. It's IMHO confusing that Simple Calendar then switches to the current date (today) and does not stay in the week of the current, selected day.

Happens to me quite often.

Aga-C:
Hi,

I've added keeping the selected date when changing views:
- Closes #1403 
- It keeps the date when we are moving to less-detailed views, e.g. from daily to weekly, from weekly to monthly. When going to more detailed views, it shows current date.
- To do this improvement correctly, I had to modify and move the logic of counting week's DateTime - now it works for any date and is moved to Context.

**Before:**

https://user-images.githubusercontent.com/85929121/146649839-19701963-fe83-4156-86cb-4f32699da339.mp4

**After:**

https://user-images.githubusercontent.com/85929121/146649842-2d38ea7d-f039-44fa-ad95-406e3a020d3b.mp4


tibbi:
should be good, just formatted some code to make it consistent with the rest. Thanks

PR #1583, opened 2021-12-18
Fixed showing unnecessary back arrow

Aga-C:
Hi,

I've found the bug that the back arrow is not hidden after changing the view when previously we were in a view opened directly from another (like in a daily view opened by clicking on a date in monthly view). The back arrow was misleading since it was closing the app instead of going back in the history.

I've fixed it by hiding the back arrow, since the logic of changing views doesn't keep the history of what has been opened.

**Before:**

https://user-images.githubusercontent.com/85929121/146652648-19a29d49-89d3-4005-b667-d33960c070c4.mp4

**After:**

https://user-images.githubusercontent.com/85929121/146652650-bb9bb776-1aed-47a9-a4c5-eb84c8691eed.mp4


tibbi:
should be good, thanks

PR #1590, opened 2021-12-24
Added limiting events on the list widget

### LINKED ISSUE #1051: Ability to hide distant future events
Hello,

When using the calendar event list widget, I can't seem to find a way to hide distant future events, which may or may not be recurring.

If it's not too hard of a feature to add, would it be possible to support a setting for specifying a duration window after which future events don't show up in the event list widget?

In such a case where there are no upcoming events, a message such as, "No upcoming events." (optionally with the duration window) might be appropriate as a placeholder.

Thank you for continuing to support this app. :bowing_man:

### LINKED ISSUE #1335: [Feature Request] widget that displays only today
I'd love for the option to restrict my home screen widget to display events for only today! I tend to take life one day at a time :).

Aga-C:
Hi,

I've added an option to limit the events in the list widget by setting a max date. 
- Closes #1335, closes #1051.
- Old widgets are set to have limit of one year, same as it was before.

https://user-images.githubusercontent.com/85929121/147347066-29ca6f75-f418-4184-bfbc-b0963e3b4a87.mp4

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/WidgetListConfigureActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/WidgetListConfigureActivity.kt
@@ -128,7 +128,9 @@ class WidgetListConfigureActivity : SimpleActivity() {
         items.add(RadioItem(EVENT_PERIOD_ONE_YEAR, getString(R.string.within_the_next_one_year)))
         items.add(RadioItem(EVENT_PERIOD_CUSTOM, getString(R.string.within_the_next)))
 
-        RadioGroupDialog(this, items, selectedPeriodOption, showOKButton = true, cancelCallback = null) {
+        val selectedOption = if (items.any { it.id === selectedPeriodOption }) selectedPeriodOption else EVENT_PERIOD_CUSTOM
+
+        RadioGroupDialog(this, items, selectedOption, showOKButton = true, cancelCallback = null) {
             val option = it as Int
             if (option == EVENT_PERIOD_CUSTOM) {
                 CustomPeriodPickerDialog(this, selectedPeriodValue, selectedPeriodValueType) { value: Int, type: Int ->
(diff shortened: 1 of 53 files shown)


app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/WidgetListConfigureActivity.kt
```diff
@@ -88,6 +104,60 @@ class WidgetListConfigureActivity : SimpleActivity() {
...
+        }
+    }
+
+    private fun showPeriodSelector() {
+        hideKeyboard()
+
+        val items = ArrayList<RadioItem>()
+        items.add(RadioItem(EVENT_PERIOD_ONE_YEAR, getString(R.string.within_the_next_one_year)))
+        items.add(RadioItem(EVENT_PERIOD_TODAY, getString(R.string.today_only)))
+        items.add(RadioItem(EVENT_PERIOD_CUSTOM, getString(R.string.within_the_next)))
```
tibbi, line 125: lets add a couple more predefined values, like 1 week, 1 month, 1 year. Sort them like this too, make Today the first one and Custom the last one.

Aga-C:
obraz

app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/WidgetListConfigureActivity.kt
tibbi: I mean predefined ones so that you dont even have to click on Custom, but they will be above it

Aga-C: I did it, it's added in a line `items.addAll(getCustomRadioItems())`. Please see screenshot I've added after last commit.

tibbi: ah right, this is done in a really weird way though.. You add some manually, some via the new function. Make it consistent please

Aga-C: I did it like this, because I'm reusing these values later in the `updateSelectedPeriod` to get text easily without a need to recalculate everything again.

tibbi: it is still pretty bad, just do it like in other apps please. You dont need to recalculate anything.

tibbi:
yeah thanks, should be good now. It looks a lot more consistent and familiar to other apps now

PR #1596, opened 2021-12-26
Update strings.xml

10cents:
I have avoided some strings, which meaning I don't understand:
In the section List widget config (line 230) you have strings like
Within the next %d day
Within the next %d days
Why 'Within the next %d day', isn't %d a placeholder for the number one? Can't I just translate it as 'Within the next day'?

### DIFF
--- a/app/src/main/res/values-da/strings.xml
+++ b/app/src/main/res/values-da/strings.xml
@@ -88,14 +88,14 @@
     <string name="no_birthdays">Der blev ikke fundet nogen fødselsdage</string>
     <string name="no_new_birthdays">Der blev ikke fundet nogen nye fødselsdage</string>
     <string name="birthdays_added">Fødselsdage tilføjet</string>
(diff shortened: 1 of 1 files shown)


Aga-C:
I also omitted it in Polish translation. I think that you should translate as it works the best for your language.

tibbi:
nope, it shouldnt be avoided. Android Studio will complain about the missing %d

10cents:
ok, thanks for clarifying it

tibbi:
so guess this isnt needed anymore

10cents:
If it is the incomplete translation, yes. I could not figure out if it 
was possible to edit it, so I started a new, hm, patch/file/attempt?

Den 29.12.2021 kl. 14.47 skrev Tibor Kaputa:

tibbi:
it is difficult, just starting a new one is easier

PR #1616, opened 2022-01-11
fix issue 1588 overlapp issue by adding crosswise collision checks

Honk2:
Do a crosswise collision check with all events of a day


tibbi:
that's impressive, I tested a few cases and found no bugs, we will see in real world how it works :) Thanks

tibbi:
@Honk2 so this seems to fail for short events, like 17:00 - 17:00. If you create event One from 17:00 to 17:00 and Two from 17:00 to 17:00, only Two will be shown. Any quick fix for it? It can be till 17:10 too.

tibbi:
or nevermind, it was caused by Range file extension, history at https://github.com/SimpleMobileTools/Simple-Calendar/commits/master/app/src/main/kotlin/com/simplemobiletools/calendar/pro/extensions/Range.kt

PR #1655, opened 2022-02-28
Fixed crash while creating monthly widget

Aga-C:
Hi,

I've found that it was impossible to create a monthly widget. Here is a fix for it.

BTW. These new colors (for past events and next month events) on monthly widget are too bright, it would look better if they would be the same as in the app.

**Before:**

https://user-images.githubusercontent.com/85929121/155964464-59d74cc9-ac70-44c5-96b5-271b7ebb00e2.mp4

**After:**

https://user-images.githubusercontent.com/85929121/155964479-07c3bd00-7d79-4b43-ba35-f3387061c913.mp4

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/extensions/Context.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/extensions/Context.kt
@@ -437,26 +437,29 @@ fun Context.addDayNumber(rawTextColor: Int, day: DayMonthly, linearLayout: Linea
     if (!day.isThisMonth)
         textColor = textColor.adjustAlpha(LOWER_ALPHA)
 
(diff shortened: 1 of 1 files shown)


tibbi:
how to reproduce the crash? What should be done?

Aga-C:
I've just updated the app on my phone. After seeing strangely bright colors on the monthly widget, I decided to remove it and add again, but it was impossible to add it. Everything's shown in `Before` video - just try to create a new monthly widget.

tibbi:
if you reinstall the app, does it still crash?

Aga-C:
Yes.

tibbi:
okay, fixed it without findViewById to make it nicer. Thanks though

Aga-C:
You could just write a comment, I'd change it in a short while.

tibbi:
sometimes it is just easier to change the code myself instead of checking what changed in the PR

PR #1669, opened 2022-03-13
Update strings.xml

10cents:


tibbi:
check lines 35+, there are many more task related strings :) Thanks

Issue #1587, opened 2021-12-23
Massive ErrorS in UK Calendar


faxotherapy:
Hello,
Good Friday for UK is 15 April on Friday only and NOT on 2 April, Saturday. Good Friday is also shown for 2 dates: 2 April and 15 April. It should not be shown twice and surely not on 2 April.

Issue 2: Easter is 18 April only and not on 2 April nor 5 April as shown on calendar…

So, the app is SimplyNotUsable if relying on those provided holiday's dates to organise trips around these dates. SimplyTooBad.

tibbi:
what app version are you using?

faxotherapy:
6.16.2

Wintandre:
Looks like Good Friday and Easter were set to annual repeating events from 2021. I have fixed these and others and have created Pull Request #1676.

PR #1676, opened 2022-03-20
Fix repeating and non-repeating UK holidays. Issue #1587.

Wintandre:
Although the app doesn't (seem to) make use of the STATUS, I've marked old holidays as CANCELLED for completeness. Some were erroneously repeating, so that has been removed and a LAST-MODIFIED date added in order that the importer will recognize the changes.

Tested importing original version, then importing this version as well as importing this version into fresh calendar. In both cases, the holidays for the next 2 years at least look OK to me.


tibbi:
looks good, thanks

Issue #1708, opened 2022-05-04
Monthly view confused by repeating events spanning midnight


Loirooriol:
Add an event like:

 - Start: May 9 at 11:00 PM
 - End: May 10 at 01:00 AM
 - Repeat: daily, 2 times

That is:
 - 1st occurrence: May 9 11:00 PM - May 10 at 01:00 AM
 - 2nd occurrence: May 10 11:00 PM - May 11 at 01:00 AM

The weekly view shows them right at the top (though they don't appear in the grid, maybe #1662):
![calendar-weekly](https://user-images.githubusercontent.com/7477678/166813737-087c944e-39e3-409d-bb9a-b2e8bf8473ed.png)

But the monthly view is wrong, it looks like the 2nd occurrence is May 11 11:00 PM - May 12 at 01:00 AM:
![calendar-monthly](https://user-images.githubusercontent.com/7477678/166813743-beb4bc48-e3cd-4cd4-90d2-4be07761f398.png)

Loirooriol:
It's version 6.18.1. This is the .ics, I guess you may need to adjust the times so that the events cross midnight in your local time, I'm in GMT+2. 

```
BEGIN:VCALENDAR
PRODID:-//Simple Mobile Tools//NONSGML Event Calendar//EN
VERSION:2.0
BEGIN:VEVENT
SUMMARY:Test
UID:1c929ad68ed0499082b4c5debe49d3a31651694478424
X-SMT-CATEGORY-COLOR:-689152
CATEGORIES:Regular event
LAST-MODIFIED:20220504T200608Z
LOCATION:
TRANSP:OPAQUE
DTSTART:20220510T210000Z
DTEND:20220510T230000Z
X-SMT-MISSING-YEAR:0
DTSTAMP:20220504T212640Z
STATUS:CONFIRMED
RRULE:FREQ=DAILY;INTERVAL=1;COUNT=2
DESCRIPTION:
BEGIN:VALARM
DESCRIPTION:Recordatori
ACTION:DISPLAY
TRIGGER:-P0DT0H10M0S
END:VALARM
END:VEVENT
END:VCALENDAR
```

PR #1722, opened 2022-05-21
Properly handle repeating events

Naveen3Singh:
Fixes https://github.com/SimpleMobileTools/Simple-Calendar/issues/1708

Repeating events spanning midnight without glitches:

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/views/MonthView.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/views/MonthView.kt
@@ -115,16 +115,17 @@ class MonthView(context: Context, attrs: AttributeSet, defStyle: Int) : View(con
     }
 
     private fun groupAllEvents() {
-        days.forEach {
-            val day = it
-            day.dayEvents.forEach {
-                val event = it
+        days.forEach { day ->
 
+            day.dayEvents.forEach { event ->
                 // make sure we properly handle events lasting multiple days and repeating ones
                 val lastEvent = allEvents.lastOrNull { it.id == event.id }
                 val daysCnt = getEventLastingDaysCount(event)
                 val validDayEvent = isDayValid(event, day.code)
-                if ((lastEvent == null || lastEvent.startDayIndex + daysCnt <= day.indexOnMonthView) && !validDayEvent) {
+
+                val lastDay = { findLastDay(event) ?: day }
+
+                if ((lastEvent == null || lastEvent.startDayIndex + daysCnt <= lastDay().indexOnMonthView) && !validDayEvent) {
                     val monthViewEvent = MonthViewEvent(
                         event.id!!, event.title, event.startTS, event.endTS, event.color, day.indexOnMonthView,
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/simplemobiletools/calendar/pro/views/MonthView.kt
```diff
@@ -429,6 +433,11 @@ class MonthView(context: Context, attrs: AttributeSet, defStyle: Int) : View(con
         return event.startTS != event.endTS && Formatter.getDateTimeFromTS(event.endTS) == Formatter.getDateTimeFromTS(date.seconds()).withTimeAtStartOfDay()
     }
 
+    private fun findLastDay(event: Event) =
```
tibbi, line 436: this is hard to read, type out the return value like (event: Event) : Day {... return.... please

app/src/main/kotlin/com/simplemobiletools/calendar/pro/views/MonthView.kt
```diff
@@ -113,16 +115,17 @@ class MonthView(context: Context, attrs: AttributeSet, defStyle: Int) : View(con
...
+        days.forEach { day ->
 
+            day.dayEvents.forEach { event ->
                 // make sure we properly handle events lasting multiple days and repeating ones
                 val lastEvent = allEvents.lastOrNull { it.id == event.id }
                 val daysCnt = getEventLastingDaysCount(event)
                 val validDayEvent = isDayValid(event, day.code)
-                if ((lastEvent == null || lastEvent.startDayIndex + daysCnt <= day.indexOnMonthView) && !validDayEvent) {
+
+                val lastDay = { findLastDay(event) ?: day }
```
tibbi, line 126: do we need the brackets here?

tibbi: ah, this is supposed to be a function? Make it a proper function then please, this is unreadable

Naveen3Singh: Yes, it's a lambda

Without the lambda, the `findLastDay()` (iterates through the whole `days` array and events) will be executed needlessly in each loop even when there are no repeating events 

I created the lambda for readability only, the `findLastDay() :? day` term could be moved to the if statement itself avoiding the need for a lambda

How about local functions? Allowed?

tibbi: that sounds really bad, just add some casual function please

tibbi:
yeah thanks, looks much better now

app/src/main/kotlin/com/simplemobiletools/calendar/pro/views/MonthView.kt
Naveen3Singh: https://github.com/SimpleMobileTools/Simple-Calendar/pull/1722/commits/ac55c060d51f53723ae43fad7955d7cd2cfe6675

(the nullability check was unnecessary)

PR #1731, opened 2022-05-31
Properly handle repeating and overlapping events

Naveen3Singh:
This pull request addresses https://github.com/SimpleMobileTools/Simple-Calendar/issues/1708

Same goal as https://github.com/SimpleMobileTools/Simple-Calendar/pull/1722 but without any side effects.


tibbi:
alright, seems good to me :) Thanks

PR #1724, opened 2022-05-25
Weekly view slider glitch fix

Naveen3Singh:
Fixes https://github.com/SimpleMobileTools/Simple-Calendar/issues/1700

**Before:**

**After:**

Issue #1700, opened 2022-04-24
All day event block not aligned


aavsss:
Hi everyone,
 When I change the slider from 7 days to 8 days, the event names are not aligned together. I have attached images below. 
I use version 6.18.1 Pro. Any recommendation on this issue is greatly appreciated. 

Thanks for help.
![Screenshot_20220424-162730](https://user-images.githubusercontent.com/48468560/164998253-8ca30007-f634-4538-a1a4-e8c2eb41340a.jpg)
![Screenshot_20220424-164619](https://user-images.githubusercontent.com/48468560/164998254-24a4cb46-89e9-43d0-8c80-70d0944da5e1.jpg)
![Screenshot_20220424-164657](https://user-images.githubusercontent.com/48468560/164998255-30ebe6d8-8e52-4e48-84d2-a4d965196432.jpg)

PR #1737, opened 2022-06-08
Avoid creating new rows unnecessarily

Naveen3Singh:
Related to bug https://github.com/SimpleMobileTools/Simple-Calendar/issues/1700 fixed in PR https://github.com/SimpleMobileTools/Simple-Calendar/pull/1724

Before this change, each repeating event was created on a different set of rows even when there was enough space for the events to fit. 

**Before vs After:**


tibbi:
as usually, it works well in my tests, but the real test will be done by the users :) Thanks

Issue #1660, opened 2022-03-04
Tasks reminders


L-Med:
Good day, this is a feature request for tasks to have the same reminder functionality as events. Thanks in advance.

tibbi:
will be added in 6.19.0

PR #1760, opened 2022-06-28
Add Task reminders

Naveen3Singh:
PR addressing https://github.com/SimpleMobileTools/Simple-Calendar/issues/1660

Users can now add reminders in tasks:
    

**Notes:**  

- Notifications are still shown for **already completed tasks** that are set to occur in the future
- "Mark completed" action won't show up if the task is already complete.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/services/MarkCompleteService.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/services/MarkCompleteService.kt
@@ -1,0 +1,30 @@
+package com.simplemobiletools.calendar.pro.services
+
+import android.app.IntentService
+import android.content.Intent
+import com.simplemobiletools.calendar.pro.extensions.cancelNotification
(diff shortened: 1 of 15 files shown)


app/src/main/kotlin/com/simplemobiletools/calendar/pro/services/MarkCompleteService.kt
```diff
@@ -0,0 +1,30 @@
...
+                task.flags = task.flags or FLAG_TASK_COMPLETED
+                eventsDB.updateTaskCompletion(task.id!!, task.flags)
+                cancelPendingIntent(task.id!!)
+                cancelNotification(task.id!!)
+            }
+        }
+    }
+
+    companion object {
+        const val ACTION_MARK_COMPLETE = "ACTION_MARK_COMPLETE"
```
tibbi, line 28: you should add such things into constants.kt

tibbi:
seems to work well, it is just missing a "d" at the end at functions etc. It is called "Mark completed", not "Mark complete"

tibbi:
alright thanks, looks good

PR #1785, opened 2022-07-20
Add task repetition

Naveen3Singh:
### Changes:
- All completed tasks are stored in a new `tasks` table
- Add save confirmation dialog on back press


tibbi:
yeah good job, looks good to me, will see if there will be some issues :)

Issue #1658, opened 2022-03-04
Entries are doubled 


topcaser:
Hi, my appointments which are synced via davx5 are shown twice. Etar calendar does not have this problem. Any idea how to fix it?

Thanks for your support.

tibbi:
what app and OS version are you using?

topcaser:
App Version 6.17.2
OS: color OS 11 based on android 11

The app is loaded from fdroid

tibbi:
you mean all events, or are those birthdays only? If you trigger some manual syncing, doesnt it create even more copies?

topcaser:
The Birthdays are only once there. But the appointments of my main davx5 synced calender which is hosted in the nextcloud are shown twice (doubled). This is not the case with Etar calendar which means they (the appointments) are there only once and the bug is induced from simple calendar

Oclair:
![Screenshot_20220310-173203_Calendar](https://user-images.githubusercontent.com/619643/157709951-816c6273-f34f-4ee7-8bdc-d0c9fa904563.jpg)

Oclair:
seems to occur when an event on the app cannot be written to the nextcloud calendar db

gg-gg1:
Hi. I have the same issue with contacts birthdays? It duplicates contacts birthdays and on some contacts "triplicates". I am not sure if this is the issue, but I think it does that because it sync with all accounts from my device. If I open the calendar app and go to settings-add contact birthdays, it adds the birthdays, but after I open Whatsapp or another app that uses contacts and I go back in the calendar app, it doubles the contacts birthdays. Hope that I explained it well.

Naveen3Singh:
@topcaser Can you pls export and attach the `.ics` file of the duplicated events here? 
That would really help in debugging this issue. Thanks

How to:
 Tap on the three dots in the top-right corner => tap on "Export events to an .ics file" => Choose storage location for export

topcaser:
Hi, I can't upload all my private events. Sorry. But I can test a possible fix.

EDavidF:
I am having this same problem - Version 6.16.2 Pro - Android 11 - ArrowOS v11.0

vista-narvas:
hi
i have the same problem
only some events are duplicated
i am not able to produce a event that duplicates
but i can share a .ics file that has one (i removed all events except the related one and changed the name)
steps i took to reproduce
[...]

Issue #1601, opened 2021-12-30
Multiple/duplicated instances of events  appearing


maze88:
May be related to recurring events and/or changes being made (remotely?) to them.

See screenshots below, comparing...

## Simple Calendar

![Screenshot_20211230-125439](https://user-images.githubusercontent.com/16841581/147764891-cae4bb90-f186-495b-ba43-8f677cfef1fe.png)

## Google calendar

![Screenshot_20211230-125841](https://user-images.githubusercontent.com/16841581/147765008-3dcb3d80-d349-4ed4-b579-dfeee277a736.png)

tibbi:
What type of event is it? Some recurring one with exceptions, maybe it has been modified? Can you somehow reproduce the issue?

benedetto:
Same over here. I have at least three different types where this occurs:
* Recurring events with exceptions
* Multi-day full-day events (not recurring)
* single, non-recurring events.

Additional information (don't know if related):
* For me, it only happens in one of several calendars
* This calendar contains (among other events) two recurring series from which I delete single events through the app on a regular basis
* Events deleted from the series above have the tendency to reappear (sometimes just reappear, sometimes as doubled events) although they are no longer on the server calendar they are synced to and from.

If I can be of any help (log files or the like...), please let me know.

tibbi:
I could use some exact steps for reproducing the issue. Are you syncing via Google or some other service?

benedetto:
I'm synching via Davx5 to a Nextcloud instance.

Unfortunately I cannot give you any steps to reproduce the behavior – I have noticed it some weeks (maybe even months) ago, but don't know for how long it already existed before or whether I did something special  beforehand.

michaelof:
I'm curious if we get this reproducible. Happens to me from time to time, for many versions of SimpleCalendar. For me never reproducible, thought sometimes bug disappeared and reintroduced a while later.
But, maybe helpful, 100% possible workaround was and is, all my calendars synched via calDAV, to remove calendar(s) from sync options, wait until events disappear, and re-add calendar(s) to be synched again.

maze88:
I'm still trying to find a way to reproduce it; glad to hear others are experiencing & trying too!

Some -stll untested- thoughts:
- Could be related to shared calendars: i.e. I have two calendars synced with the app, but between them (via gsuite) they're also shared and visible to each other - or at least one way.
[...]

PR #1800, opened 2022-08-09
Fix event duplication issues

### LINKED ISSUE #1778: "Add contact anniversaries" makes multiple event types on renaming and/or change of language
If I use the function "Add contact anniversaries" it creates a new event type. In Danish, the default name for this event type is "Mærkedage". If I change that name to something else (via Settings - General - Manage event types - <select event type> - Edit), Simple Calendar first just rename the event type, but soon after it makes a copy of the event type, so there will be two event types with same colour, but with the old (default) and the new (renamed) name. 

Also, if I don't rename anything, and I only have one event type with the default Danish name, and then select "Settings - General - Use English language", it creates a copy of the event type, so I now have two event types with same colour, but with the default Danish and default english name.

Naveen3Singh:
PR addressing issues https://github.com/SimpleMobileTools/Simple-Calendar/issues/1601, https://github.com/SimpleMobileTools/Simple-Calendar/issues/1658, and https://github.com/SimpleMobileTools/Simple-Calendar/issues/1778

### Notes
- There are cases in #1601 that I wasn't able to reproduce "organically" so it's possible that the fix doesn't actually fix the same case, we'll have to wait for the users to report.
- To fix #1778, a new column was added to the `event_types` table, and migration is needed. As there are no ways to check what event type is used for say birthday events, users with duplicated event types will have to delete them manually after updating
- All CalDAV-related testing was done using Nextcloud (trial) + DAVx5 + Thunderbird on fedora


matham:
Thank you! This fixed the duplication issue for me!

Can't wait for an official release so I can remove my manually built debug apk.

PR #1811, opened 2022-08-20
CalDAV refresh improvement

Naveen3Singh:
Synchronize local DB with android calendar DB in `onResume`


tibbi:
showing events at toggling CalDAV sync is fixed now, but recognizing events changed/added/deleted in Google Calendar still takes a long time and nothing really happens at clicking RefreshCalDAV calendars. Guess Ill apply it and see some feedback. Thanks

PR #1814, opened 2022-08-23
New Zealand .ics

floatingfoo:
Adds newzealand.ics file to source folder


tibbi:
this is just the ics, but remove the empty lines please and add it at the available countries. Have you tested importing it?

Issue #1799, opened 2022-08-08
Events spanning across midnight are not displayed in week view, if their start date is before the currently displayed week


andy-berg:
If the option "Show events spanning across midnight at the top bar" is deselected, events spanning across midnight are not displayed in week view, if their start date is before the currently displayed week.

moonyGopher:
In version 6.20.0 (from F-Droid) multi day events over multiple weeks are shown correctly in the week view. If Upgrading to version 6.20.1 (from F-Droid) the multi day events are only shown in the in the first week of the event.

Screenshots from version 6.20.0

![7-day-view_first-week_version-6-20-0](https://user-images.githubusercontent.com/88974951/195052735-7766d12f-76a4-4bac-b533-1a7b3d865fe8.png)

![7-day-view_second-week_version-6-20-0-](https://user-images.githubusercontent.com/88974951/195052741-b16c2fb4-faae-427a-9990-cc213f5348fd.png)


Screenshots from version 6.20.1

The event is only shown in the first week where the event should be.
![7-day-view_first-week](https://user-images.githubusercontent.com/88974951/195051181-b940881f-2c71-4d63-b8a0-5cb23a39aa1b.jpeg)

![7-day-view_second-week](https://user-images.githubusercontent.com/88974951/195051187-fea78a83-1d0b-4bf4-903e-88f00e8ca785.jpeg)

The event is shown if changing the week view to show all days where the event should be.
![8-day-view](https://user-images.githubusercontent.com/88974951/195051190-d0df7d00-c81c-42d8-802f-e1c3b3aecbb4.jpeg)

Issue #1767, opened 2022-07-02
Option "Show events spanning across midnight at the top bar" does not work correctly


dumblob:
As described in https://github.com/SimpleMobileTools/Simple-Calendar/issues/1241#issuecomment-1153982721 the option `Show events spanning across midnight at the top bar` does not work as if I **de-select** it, then `all-day` events are still shown at the top:

![Screenshot_20220613-162013](https://user-images.githubusercontent.com/2451044/173375519-f0855b53-bc64-4935-9718-ab2d06fbf376.jpg)

tibbi:
what events are those? I mean start/end dates and times? What app version are you using?

dumblob:
> what events are those? I mean start/end dates and times?

`all-day` set as "All day" (i.e. no start/end times), June 13 - June 18, UTC, No reminder, No repetition, Store locally only, Regular event

`non-all-day` June 13 17:00 - June 17 17:30, Europe/Amsterdam, No reminder, No repetition, Store locally only, Regular event

>What app version are you using?

Version 6.18.3 Pro

PR #1821, opened 2022-09-02
bugfix multi week event not showing in next week

DoozyDoz:
## Description
- As reported in issue #1799 , events that span more than one week were not visible in the last week.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt
@@ -537,7 +537,7 @@ class WeekFragment : Fragment(), WeeklyCalendar {
         }
 
         dayevents@ for (event in events) {
-            val startDateTime = Formatter.getDateTimeFromTS(event.startTS)
+            val startDateTime = Formatter.getDateTimeFromTS(if (weekTimestamp > event.startTS) weekTimestamp else event.startTS)
             val startDayCode = Formatter.getDayCodeFromDateTime(startDateTime)
             val endDateTime = Formatter.getDateTimeFromTS(event.endTS)
             val endDayCode = Formatter.getDayCodeFromDateTime(endDateTime)


tibbi:
could you give exact steps for reproducing what does this fix? With specific dates to be set.

DoozyDoz:
> could you give exact steps for reproducing what does this fix? With specific dates to be set.



Hello @tibbi,
Thanks for taking time to review, below are the steps to reproduce the bug and the last is the screen shot of the fix 
<p float="right">
<em>1. First uncheck this option in settings -> weekly view</em><br>

<br>
<em>2. Switch to week view</em><br>
<br>
<em>3. Create an event that spans two weeks eg. </em><br>
  <br>
<em>4. [BUG] The event will not appear in the second week only the first week. </em><br>
<br>
<em>5. [FIX] The event appears as expected. </em><br>

</p>

dumblob:
Yep, the option `Show events spanning across midnight at the top bar` seems to have also other buggy behavior - e.g. https://github.com/SimpleMobileTools/Simple-Calendar/issues/1767 .

DoozyDoz:
> Yep, the option `Show events spanning across midnight at the top bar` seems to have also other buggy behavior - e.g. #1767 .

@dumblob , [PR 1837](https://github.com/SimpleMobileTools/Simple-Calendar/pull/1837) should fix this, kindly test on your side too

dumblob:
@DoozyDoz wow, that fix looks good!

By any chance, do you or @tibbi have any CI regularly creating `apk`s for PRs? I do not have Android studio available now so I would need a binary to test it. So far I noticed @Naveen3Singh is preparing such binaries only manually.

DoozyDoz:
@tibbi did you forget this repo?

tibbi:
no, Im just always hesitant with such PRs :) They always fix one case but break something else.

dumblob:
I have tested it quickly and it seems to work for me (but of course, there might be some edge cases which my regular calendars do not contain/exhibit).

Could you @tibbi also test it? Thanks!

tibbi:
Im getting a null pointer exception at line 574: y = currentEventWeeklyView!!.range.lower * minuteHeight

tibbi:
@DoozyDoz working on the crash?

DoozyDoz:
@tibbi oh i had missed this, let me fix it

DoozyDoz:
Hello @tibbi
I have failed to reproduce the null pointer error, may I ask in which situation you got it?

tibbi:
I have an event from October 21, 23:00 to October 23, 23:00 and when I swipe to it, it crashes. Try creating random long events

DoozyDoz:
I have tried creating different long events (some spanning 2 weeks, others 3 weeks) I only get the error if i uncheck "Start week on Sunday" in settings

Naveen3Singh:
Quick look that might help:

It crashes probably because the view was stored with a different key in the first loop, see https://github.com/SimpleMobileTools/Simple-Calendar/blob/af84b9a5af8288d825a569b8db24d248bcd67cc9/app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt#L451-L461

And `startDayCode` at line 574 now has a different day code after the proposed changes than the one on line 451. Both should be same.

PR #1827, opened 2022-09-04
add german local reoccurring holidays

anoymouserver:
Since 2019 there are these 2 new yearly holidays:
- Internationaler Frauentag (Berlin)
- Weltkindertag (Thuringia)


tibbi:
looks good, thanks

PR #1830, opened 2022-09-06
Correct behaviour when fallback to other languages

susosl:
Currently, when the primary system language isn't supported by this app, it directly fallbacks to English, instead of continue checking the rest of the languages configured in the system's language settings. This can be solved by adding the resConfigs option, that prevent libraries adding strings to the build in languages the app doesn't support.


tibbi:
that doesnt look like a good idea

Issue #1798, opened 2022-08-06
Show duration of events that spans multiple days


mueller-ma:
When an event spans multiple days, the duration could be displayed in the day view, e.g. day 1 of 7:

````
Holiday 1/7
Allday
Some location
````

tibbi:
what app version and which view are you using?

mueller-ma:
I'm using version 6.19.0 Pro and request this feature for the day view.

tibbi:
if the start or end day isnt the currently shown one, it is written at the start/end time. I dont plan changing it to 1/7 or so

mueller-ma:
I created a test event: All day from 3rd September to 4th October.
This is how in looks in day view:
![Screenshot_20220808-180114_Kalender](https://user-images.githubusercontent.com/22525368/183462426-79ddb40e-96c7-4628-bb74-d9e1634a06b0.png)

and this is the event list:
![Screenshot_20220808-180212_Kalender](https://user-images.githubusercontent.com/22525368/183462428-4062ae62-b234-49ff-bea8-f394909d0f36.png)


> if the start or end day isnt the currently shown one, it is written at the start/end time

Do I have to enable this somehow?

Also note that "4 Dienstag" ("4 Tuesday") doesn't mention the month, so I have no idea how long the event goes, if I don't want to change the view.

PR #1836, opened 2022-09-11
fix issue #1798 display of start/end for multi day events

Honk2:


tibbi:
what does this fix?

tibbi:
what does this fix, can you give a sample?

dumblob:
Just makes the string shorter if possible for better UX.

Honk2:
* There is an event that spans over several days
* Switch to Day view
* When you scroll over the days, the event will only be shown as all day event, but not when it started or when it ends
grafik
grafik
grafik

With this PR,
* on the first day the end date/time is shown
* on the last day the start date/time is shown
* on the other days, the start and end date/time is shown.
* The date is shown as [month] [day]
grafik
grafik
grafik

AS described in issue #1798

tibbi:
should be good, thanks

Issue #1539, opened 2021-11-12
All day events are created on the correct day, but are moved to the next day after sync.


tnyeanderson:
Thank you for your very hard work on this project. I've used it for years and have noticed an issue that has been occurring for a long time.

Syncing via DAVx5 to Nextcloud.

Create an event on this app, set for all day... for instance on 12/7. When exported, you see it looks correct:
```
BEGIN:VCALENDAR
PRODID:-//Simple Mobile Tools//NONSGML Event Calendar//EN
VERSION:2.0
BEGIN:VEVENT
SUMMARY:Tues 7 test
UID:Caldav-3-2603
X-SMT-CATEGORY-COLOR:-13534004
CATEGORIES:Personal
LAST-MODIFIED:20211111T234155Z
LOCATION:
TRANSP:OPAQUE
DTSTART;VALUE=DATE:20211207
DTEND;VALUE=DATE:20211208
X-SMT-MISSING-YEAR:0
DTSTAMP:20211111T234158Z
STATUS:CONFIRMED
DESCRIPTION:
END:VEVENT
END:VCALENDAR
```

The event is displayed (normally) on the day it was created until DAVx5 performs a sync operation. At this point, in all clients the event is moved to the day after its originally scheduled date (moved from tues to wed in this case). Export shows that DTSTART and DTEND are now one day later each:
```
BEGIN:VCALENDAR
PRODID:-//Simple Mobile Tools//NONSGML Event Calendar//EN
VERSION:2.0
BEGIN:VEVENT
SUMMARY:Tues 7 test
UID:Caldav-3-2603
X-SMT-CATEGORY-COLOR:-13534004
CATEGORIES:Personal
LAST-MODIFIED:19700101T000000Z
LOCATION:
TRANSP:OPAQUE
DTSTART;VALUE=DATE:20211208
DTEND;VALUE=DATE:20211209
X-SMT-MISSING-YEAR:0
DTSTAMP:20211111T234338Z
STATUS:CONFIRMED
DESCRIPTION:
END:VEVENT
END:VCALENDAR

```

As an example, I downloaded Etar, which handles all day events differently. The way Etar handles it works great in all clients before and after sync. Exports from the Etar app of events created in Etar are shown below.

Before sync:
```
BEGIN:VCALENDAR
VERSION:2.0
CALSCALE:GREGORIAN
METHOD:REQUEST
PRODID:-//Etar//ws.xsoh.etar
BEGIN:VEVENT
[...]

Issue #1521, opened 2021-10-23
Importing ics file as default entries shifts all dates backward 1 day


Sepero:
Reproducibility 100%

Actions to Reproduce:
1. Export all entries
2. Delete all entries
3. Reimport all entries as CalDev (ignoring event types in the file and use the default)
4. All entries appear correct at first, but leave the app and come back, and they are all shifted backward in time by 1 day

Important Note: If I reimport the entries without using CalDev (retain event types in the file), then the import is correct and dates are Not shifted.

This bug may be related to #1480 . My timezone is also in Asia, and all of my events are all-day events.

Android v10
Simple Calendar v6.15.4

Naveen3Singh:
@Sepero 

If you are willing to test, here is a test APK with a possible fix: https://github.com/Naveen3Singh/Simple-Calendar/releases/download/all_day_event_fix/calendar-prepaid-debug.apk ([info](https://github.com/Naveen3Singh/Simple-Calendar/releases/tag/all_day_event_fix))

APK is for testing only and should be uninstalled afterward.

Lmk if it works (or not)

Thx much.

Sepero:
I'll try to make time to test it. Thank you very much 🙂👍🏻

Issue #1480, opened 2021-09-22
All-day events sometimes registered one day behind.


NaofumiHonda:
My timezone is Asia/Tokyo and my caldav server is davical through davx5.

Usualy all-day events are registered in the server correctly, but
when the registration of an all-day event is done from about 23 PM to 1 AM(GMT+9), 
they are registered in the server with their event dates being one day behind.

If UTC timezone is selected in the simple calendar app, then every thing goes well.
Furthermore, in my investigation, other clients like thunderbird lightning, evolution calendar, etc
use the UTC timezone for all-day events (of course, they use a local timezone for other type of events).

I am now thinking that the UTC timezone is better to be selected as a default timezone for all-day events
even if the simple calendar app. Or such a setting will exist.

tibbi:
what app version are you using?

NaofumiHonda:
The most recent pro version: v6.15.1

Naveen3Singh:
@NaofumiHonda 

If you are willing to test, here is a test APK with a possible fix: https://github.com/Naveen3Singh/Simple-Calendar/releases/download/all_day_event_fix/calendar-prepaid-debug.apk ([info](https://github.com/Naveen3Singh/Simple-Calendar/releases/tag/all_day_event_fix))

APK is for testing only and should be uninstalled afterward.

Lmk if it works (or not)

Thanks.

tibbi:
should be fixed in 6.20.2 released soon

Issue #1256, opened 2020-12-27
Time Zone Messes Up Date


frink:
v6.11.3 Pro

I've noticed that when the timezone overlaps days it puts it on the wrong day. Like say you're in US and meeting with someone from Australia. Their Wednesday morning is your Tuesday evening. But the Event will show on your Wednesday, their Thusday. So I think it's grabbing the day pre-timezone conversion while using the time post timezone conversion.

This may be what is happening in #1249 and others. Not sure.

frink:
This is from an imported calendar.

If I created an event locally or move the event it does not seem to have this problem...So there is definitely something wonky.

tleydxdy:
This happens on locally created events as well. e.g. add a recurring event on may 10 (mon) 19:00-19:30 America/Toronto timezone, and view it in Asia/Shanghai you will see the event is completely messed up. It puts the first event on next week's monday (should be tuesday).

Eqfour:
Hello,
I got the same issue like in #1466 
I was really confued because all of my yearly reocurring all day events (birthdays) are moved to one day before as I traveld from Europe to South America.. 
Is there any fix yet?

Weirdly the only thing that stayed in place is my own birthday.

Thanks a lot for your help and thank you SO much for all of your Apps!!
Cheers

inson1:
@Eqfour SMT apps were sold to ad company. Fork is here https://github.com/FossifyOrg

Eqfour:
Oh damn. I wasn't aware of that. Thank you so much for letting me know!

PR #1862, opened 2022-10-20
All-day event timezone fixes

Naveen3Singh:
Closes issues https://github.com/SimpleMobileTools/Simple-Calendar/issues/1539, https://github.com/SimpleMobileTools/Simple-Calendar/issues/1480, https://github.com/SimpleMobileTools/Simple-Calendar/issues/1521

### Notes: 
- All-day events now use UTC timezone
- It's no longer possible to change the timezone when creating or editing an all-day event (until All day checkbox is unticked)
- When importing an all-day event from an ICS file, the start and end timestamps now are offset by hour  13 instead of 5 (as per [this line](https://github.com/SimpleMobileTools/Simple-Calendar/blob/4aa78bfb0bc858d3941458e5b269f770c7b664f3/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/Formatter.kt#L138)), this fixes the issue in #1521. The bug was likely introduced in commit https://github.com/SimpleMobileTools/Simple-Calendar/commit/37c86400d4dfc80138a9a8f6eefae0097e191a9e
- #1256 and #1249 are related issues but not addressed in this PR

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/EventActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/EventActivity.kt
@@ -224,9 +224,14 @@ class EventActivity : SimpleActivity() {
         updateTextColors(event_scrollview)
         updateIconColors()
         refreshMenuItems()
-        event_time_zone_divider.beVisibleIf(config.allowChangingTimeZones)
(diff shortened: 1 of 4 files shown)


tibbi:
hard to test now, hope it will be ok :) Thanks

PR #1864, opened 2022-10-23
chore(holidays): australia & malaysia

curbengh:
replace 2021 dates with 2023's


app/src/main/assets/malaysia.ics
```diff
@@ -320,45 +321,60 @@ RRULE:FREQ=YEARLY;INTERVAL=1
...
 BEGIN:VEVENT
-DTSTART;VALUE=DATE:20210719
-DTEND;VALUE=DATE:20210720
+DTSTART;VALUE=DATE:20230627
+DTEND;VALUE=DATE:20230628
 UID:mal_59
 STATUS:CONFIRMED
 SUMMARY:Arafat Day (Terengganu)
 END:VEVENT
 END:VCALENDAR
```
tibbi, line 365: the file has to end with this END:VCALENDAR and not END:VEVENT

tibbi:
have you tested if this works at all?

app/src/main/assets/malaysia.ics
curbengh: fixed

curbengh:
tested on v6.20.1.228

PR #1869, opened 2022-10-31
Properly handle all-day event timezone

Naveen3Singh:
Fixes issues like all-day events moving +-1 day when synchronized

Because of the way timezones are handled at the UI level, all-day events are shifted to the local timezone so that they are displayed on the proper date and shifted back to UTC when exporting them.

_shifted as in [withZoneRetainFields](https://joda-time.sourceforge.net/apidocs/org/joda/time/DateTime.html#withZoneRetainFields(org.joda.time.DateTimeZone))_


tibbi:
events show wrong on the weekly view now. Instead of being at the top All-day bar, they show up as 10 minute events from 00:00 to 00:10

tibbi:
only with disabled "Show events spanning across midnight at the top bar".

tibbi:
actually it wasnt broken in this PR, master shows it wrong too

Naveen3Singh:
:)

PR #1875, opened 2022-11-07
Ignore stored hour for all-day events in weekly view

Naveen3Singh:


dumblob:
Yes, this makes all-day events again an appealing option! Thanks a lot!

tibbi:
should be good, we will see if anything breaks ;) Thanks

Issue #1863, opened 2022-10-22
Export events to an .ics file does not follow RFC 5545 formatting guidelines


Para-lyzed:
Hello, I was having trouble importing an .ics file that was exported from Simple Calendar to another program, so I ran a validation check on https://icalendar.org/validator.html to see what the issue was. The .ics file exported from Simple Calendar appeared to use the LF line break, whereas the standard for .ics files as per [RFC 5545](https://icalendar.org/iCalendar-RFC-5545/3-1-content-lines.html) is CRLF line breaks. This caused the error, and converting all the LF breaks to CRLF using [this tool](https://app.execeratics.com/LFandCRLFonline/?l=en) fixed the issue and allowed the new .ics file to pass the iCalendar verification, as well as become importable in the aforementioned program that would not previously accept the file.

Additionally, while I know less in this area, I believe .ics files should *ideally* not have empty fields, but I found that every single event in the .ics file has an empty "DESCRIPTION:" and "LOCATION:" field (since I never use them when setting an event). This is also a potential compatibility issue, because it's outside the standard, and has caused issues with certain data types before (see [kewisch/ical.js#462](https://github.com/kewisch/ical.js/issues/462#issuecomment-776623539)). I have not personally seen either of these types causing problems when empty, but excluding them is the best option for compatibility. If a field is left blank by the user, it should not be included in the .ics file.

I believe this should hopefully be a very simple fix, and I'm surprised to see that it is not currently in the open issues.

My device details are below, but I highly doubt this is a device specific issue:

OnePlus 7 Pro
Android 11
LineageOS 18 (build 18.1-20221016-nightly-guacamole)

tibbi:
should be fixed in 6.20.2

PR #1876, opened 2022-11-09
Use the correct line seperator as defined in RFC 5545 formatting guidelines, which can otherwise lead to an error

Chris-GW:
As described in Issue #1863, this fixes a formatting error in the .ics export. Without this change, the [ical validation](https://icalendar.org/validator.html) fails and problems may occur during import.

- Use CRLF sequence as line separator string "\r\n" in IcsExporter as defined in formatting guidelines [RFC 5545 - 3.1. Content Lines](https://icalendar.org/iCalendar-RFC-5545/3-1-content-lines.html).
- Omitting the export of empty LOCATION and DESCRIPTION properties in IcsExporter, because they are [OPTIONAL](https://icalendar.org/iCalendar-RFC-5545/3-6-1-event-component.html).

icalendar validator error

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/IcsExporter.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/IcsExporter.kt
@@ -15,6 +15,7 @@ import com.simplemobiletools.commons.extensions.writeLn
 import com.simplemobiletools.commons.helpers.ensureBackgroundThread
 import java.io.BufferedWriter
 import java.io.OutputStream
+import java.io.OutputStreamWriter
 
 class IcsExporter {
     enum class ExportResult {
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/IcsExporter.kt
```diff
@@ -47,7 +48,20 @@ class IcsExporter {
                 activity.toast(R.string.exporting)
             }
 
-            outputStream.bufferedWriter().use { out ->
+
+            object : BufferedWriter(OutputStreamWriter(outputStream, Charsets.UTF_8)) {
```
Chris-GW, line 52: Unfortunately I did not find another solution without overriding the `newLine()` method, as I did not want to change the `System.lineSeparator`.

app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/IcsExporter.kt
```diff
@@ -47,7 +48,20 @@ class IcsExporter {
...
+                /**
+                 * Writes a line separator. The line separator string is defined by RFC 5545 in 3.1. Content Lines:
+                 * Content Lines are delimited by a line break, which is a CRLF sequence (CR character followed by LF character).
+                 *
+                 * @see <a href="https://icalendar.org/iCalendar-RFC-5545/3-1-content-lines.html">RFC 5545 - 3.1. Content Lines</a>
+                 */
+                override fun newLine() {
+                    write(lineSeparator)
+                }
+            }.use { out ->
```
Chris-GW, line 64: Unfortunately I did not find another solution without overriding the `newLine()` method, as I did not want to change the `System.lineSeparator`.

tibbi:
looks a bit cumbersome indeed, but hope it will work well. Thanks

Issue #1847, opened 2022-09-29
Regression: CalDAV-Syncs with DAVx5 are always "manual"


593helmut:
This is a Regression that was introduced with Version 6.20.0. Issue #1150 is back:
DAVx5 always complains about Server not available if WIFI sync restriction is set in DAVx5 and the WIFI is not available.

tibbi:
this should be fixed already

Issue #1150, opened 2020-05-27
CalDAV syncs are always flagged "manual", causing issue with DAVx5


593helmut:
I use DAVx5 to sync with my local CalDAV server. DAVx5 can be set to sync only on WLAN and only if connected to a given SSID. This is useful if you have a local CalDAV server that is not reachable from the internet.
When using Simple Calendar Pro, DAVx5 always complains about connection errors with my CalDAV server if I'm not in my local WLAN. The developer of DAVx5 found that the sync is always done with the "manual" flag. This causes DAVx5 to override the WLAN and SSID restrictions. You can find my conversation with the developer of DAVx5 on https://forums.bitfire.at/topic/2285/caldav-sync-doesn-t-honor-wlan-restriction.
So please change Simple Calendar Pro to sync without the "manual" flag on normal syncs. The "manual" flag should only be used if the user explicitly forces a sync via a menu entry "sync now" or the like.

svenmoon:
I have the same issue with DAVx5 and Simple Calendar Pro.

peterDSM:
I can confirm this behavior, too with DAVx5 3.3.5 and Simple Calendar Pro 6.11.1.

593helmut:
I thought this is fixed. I actually have DAVx5 3.3.2-ose and Simple
Calendar Pro 6.10.3 and I don't have this issue anymore.

Am 04.11.20 um 18:22 schrieb peterDSM:
> I can confirm this behavior, too with DAVx5 3.3.5 and Simple Calendar Pro 6.11.1.
>

janggster:
The issue is still persisting in Calendar Pro v6.11.2 and davx5 3.3.7.

tibbi:
should be fixed in 6.13.0, let me know if it works, or if it broke anything

593helmut:
Seems to be fixed with Calendar Pro v6.13.0. Plays nicely with DAVx5 3.3.9-ose. No connection error popups on time based syncs an on syncs triggered by new or changed calendar entries.
Thumbs up!

PR #1877, opened 2022-11-09
Set the "Manual" flag in the caldav sync, only if it is triggered by the user and he is notified via toasts

Chris-GW:
As described in ticket #1847, we should not always send the `SYNC_EXTRAS_MANUAL` flag to DAVx5, instead we should only set it for user-initiated synchronisations. Otherwise, the forced synchronisations will lead to many error messages in DAVx5 when no permitted WLAN is available.

Assume that the caldav sync is `MANUAL` when we show toasts to the user as in:
- ✔️ MainActivity -> swipe reload
- ✔️ MainActivity -> MenuItem "refresh_caldav_calendars"
- ✔️ SettingsActivity -> changes in caldav calendar picker
- ❌ CalDavHelper -> insert, delete, edit caldav events
- ❌ CalDavHelper -> scheduleCalDAVSync() with alarm service every 2h

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/MainActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/MainActivity.kt
@@ -134,7 +134,7 @@ class MainActivity : SimpleActivity(), RefreshRecyclerViewListener {
         }
 
         swipe_refresh_layout.setOnRefreshListener {
-            refreshCalDAVCalendars(false)
+            refreshCalDAVCalendars(true)
         }
 
         checkIsViewIntent()
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/extensions/Context.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/extensions/Context.kt
@@ -603,7 +603,10 @@ fun Context.refreshCalDAVCalendars(ids: String, showToasts: Boolean) {
 
     Bundle().apply {
         putBoolean(ContentResolver.SYNC_EXTRAS_EXPEDITED, true)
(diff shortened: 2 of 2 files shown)


tibbi:
what exactly does this fix? Have you tested it properly?

Chris-GW:
I have only tested it on my Android device in the following setup:
- Nextcloud as CalDAV calendar source
- DAVx5 as ContentProvider with **WLAN only** enabled
- Simple calendar for viewing

With this change, the WLAN only restriction is better respected by Simple Calendar, as we don't always set the `SYNC_EXTRAS_MANUAL` flag. The presence of the `MANUAL` flag forces a synchronisation even if set restrictions are not fulfilled (as can be seen in the [code snipplet from DAVx5](https://github.com/bitfireAT/davx5-ose/blob/b17b38ff50e78c2409cede35e70ab2fdd2e56c4f/app/src/main/java/at/bitfire/davdroid/syncadapter/CalendarsSyncAdapterService.kt#L38-L47)).
Basically, Simple Calendar should not force a `MANUAL` synchronisation if it is not explicitly triggered by the user (as in the cases listed above). This is also addressed in ticket #1847 and #1150.

```kotlin
override fun sync(account: Account, extras: Bundle, authority: String, httpClient: Lazy<HttpClient>, provider: ContentProviderClient, syncResult: SyncResult) {
    try {
        val accountSettings = AccountSettings(context, account)

        /* don't run sync if
           - sync conditions (e.g. "sync only in WiFi") are not met AND
           - this is is an automatic sync (i.e. manual syncs are run regardless of sync conditions)
         */
        if (!extras.containsKey(ContentResolver.SYNC_EXTRAS_MANUAL) && !checkSyncConditions(accountSettings))
            return
       // ...
}
```

tibbi:
well ok, nothing broke during my testing, lets hope it works well :) Thanks

PR #1889, opened 2022-11-24
Always show single day all-day events at the top

Naveen3Singh:
**Before:**
The following events were shown at the top as per the user preference:
 - Any event spanning multiple days 
 - All all-day events
 
**Now:**
 - Any event spanning multiple days (all-day or not) is still shown at the top as per the user preference
 - Any all-day event lasting only one day is now always shown at the top **regardless** of user preference

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt
@@ -700,7 +700,7 @@ class WeekFragment : Fragment(), WeeklyCalendar {
     private fun shouldAddEventOnTopBar(isAllDay: Boolean, startDayCode: String, endDayCode: String): Boolean {
         val spansMultipleDays = startDayCode != endDayCode
         val isSingleDayAllDayEvent = isAllDay && !spansMultipleDays
-        return isSingleDayAllDayEvent || spansMultipleDays && config.showMidnightSpanningEventsAtTop
+        return isSingleDayAllDayEvent || (spansMultipleDays && config.showMidnightSpanningEventsAtTop)
     }
 
     @SuppressLint("NewApi")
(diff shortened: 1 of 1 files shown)


dumblob:
What is the motivation behind this move? I use the preference to not show anything at the top to save screen space. This PR though seems to undermine this option. Why does the user option exist then?

tibbi:
1 day all-day events should always be shown at the top, that is the expected behaviour

dumblob:
May I ask why is it expected?

For me it is clearly unexpected. If I choose to never show anything at the top and then suddenly something pops up there then I am truly experiencing a bug.

You are saying it is not a bug. So why does the half-working user setting to not show anything on the top even exist?

tibbi:
what setting are you referring to? Show events spanning across midnight at the top bar" does not cover 1 day all-day events

app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt
```diff
@@ -696,6 +697,12 @@ class WeekFragment : Fragment(), WeeklyCalendar {
         }
     }
 
+    private fun shouldAddEventOnTopBar(isAllDay: Boolean, startDayCode: String, endDayCode: String): Boolean {
+        val spansMultipleDays = startDayCode != endDayCode
+        val isSingleDayAllDayEvent = isAllDay && !spansMultipleDays
+        return isSingleDayAllDayEvent || spansMultipleDays && config.showMidnightSpanningEventsAtTop
```
tibbi, line 703: lets add brackets here to guarantee readable priorities

Naveen3Singh: :+1:

dumblob:
>what setting are you referring to? "Show events spanning across midnight at the top bar" does not cover 1 day all-day events

You are completely right, this is my bad reading/understanding and I am sorry for being this nudgy.

Is there any way to make all all-day events show up the same?

Now it seems 1 day all-day event will show at the top while longer all-day events will show in the days. If it is so, then it is very surprising and for me undoubtedly unwanted behavior.

tibbi:
that is exactly the intended behaviour now, will see in the future if it needs change

Issue #1715, opened 2022-05-15
Customisable font size in monthly view


michaelspunkt:
First oft all: thank you for providing such a calendar app!

I wish it was possible to change font size in the monthly view.

Pranab009:
Please if possible then add this feature.

PR #1905, opened 2022-12-13
Month view setting

mjsjml:
Customisable font size in monthly view #1715 

Added a font size setting in the settings section for Monthly view.  This is my first pul request.


tibbi:
ah thanks but no, this doesnt work well. Ill try thinking of some solution without adding an extra setting item

PR #1909, opened 2022-12-16
Update Sri Lankan holidays - Year 2023

Dilshan-H:
- Added government holidays (Public and Bank Holidays) in Sri Lanka for the year 2023
- Source data: [Department of Government Printing](http://www.documents.gov.lk/files/cdr/2023/2023_E.pdf)


tibbi:
should be good, thanks

Issue #1666, opened 2022-03-09
Export tasks as VTODO


pfertyk:
Hi!

Thank you for this app, it's really great and very useful! I've noticed, after a recent update, that I can now create tasks! That's a fantastic idea, and it really helps me organize my time ;)

However, I noticed that tasks are not exported to the .ics file. They are simply missing. I think there is a VTODO element in iCalendar format, maybe tasks could be exported that way?

Naveen3Singh:
@pfertyk @rarealphacat if you guys are willing to test, I have uploaded a test APK here with support for importing/exporting tasks: https://github.com/Naveen3Singh/Simple-Calendar/releases/download/vtodo_test_release/calendar-prepaid-debug.apk

or if you want, you can compile the APK yourself from this branch: https://github.com/Naveen3Singh/Simple-Calendar/tree/feature_export_tasks

thanks!

Aga-C:
@Naveen3Singh I've just tested it on my phone (Android 13), and it works fine 👍

PR #1937, opened 2023-01-14
Add option to export/import tasks

Naveen3Singh:
This PR addresses https://github.com/SimpleMobileTools/Simple-Calendar/issues/1666

### Changes
 - Add support for exporting/importing tasks
 - Use UUID as import ID for tasks as well
 - Make exporting events optional
 - Fixed export issue related to future repeating events (repeating events were not exported if `exportPastEvents` was false)

**Note:** Recurrences of a repeating task that were marked completed are not yet exported i.e. when the exported ICS is imported again, those specific recurrences will be marked incomplete.


app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/IcsExporter.kt
```diff
@@ -153,4 +127,67 @@ class IcsExporter {
...
+            fillDescription(event.description.replace("\n", "\\n"), writer)
+            fillReminders(event, writer, reminderLabel)
+            fillIgnoredOccurrences(event, writer)
+
+            eventsExported++
+            writeLn(END_EVENT)
+        }
+    }
+
+    private fun writeTodo(writer: BufferedWriter, task: Event) {
```
tibbi, line 163: rename it to Task to make it clear

app/src/main/res/layout/dialog_export_events.xml
```diff
@@ -48,15 +48,66 @@
...
+                android:background="@null"
+                android:checked="true"
+                android:clickable="false"
+                android:layoutDirection="rtl"
+                android:padding="@dimen/medium_margin"
+                android:text="@string/export_tasks" />
+
+        </RelativeLayout>
+
+        <ImageView
```
tibbi, line 96: add an id to all views

tibbi:
cmon, show some error message if both events and tasks are unchecked, at least No items found. Also reduce the checkbox paddings a bit, they are too high now. Make it the same as in the app settings.

tibbi:
events are also missing some UID so at export/import they will be duplicated

tibbi:
actually we have no_entries_for_exporting, so use that when events and tasks are both unchecked, so we dont have to add new strings. But show it right at pressing OK, dont allow selecting a file etc.

Naveen3Singh:
> Also reduce the checkbox paddings a bit, they are too high now. Make it the same as in the app settings.

I thought so too but the paddings are the same as app settings, it just looks taller because the width is lesser here. 
I'm reducing the vertical paddings to 4dp+4dp (from 8dp+8dp)

Naveen3Singh:
> tasks are also missing some UID so at export/import they will be duplicated

right, for some reason, I thought duplicating the events was the default behavior :smile:
 
I'll add a method to migrate existing tasks to use import Ids (for existing users who created tasks)

Naveen3Singh:
Check now please.

tibbi:
those paddings are too small now, just check how are the settings done...

Naveen3Singh:
image

it was the same as the settings. 8dp on top and 8dp on bottom.
I [just](https://github.com/SimpleMobileTools/Simple-Calendar/pull/1937/commits/0725eabcb229c835ccaee2eea65729c0d4f2509e) increased it to 6dp from 4dp.

tibbi:
the settings use 8 dp

Naveen3Singh:
Yes.
I'm saying the dialog also used 8dp when I created the PR, I decreased it to 6dp because you said it looked a bit too big in this situation.

Naveen3Singh:
Yes.
I'm saying the dialog also used 8dp when I created the PR, I decreased it to 6dp because you said it looked a bit too big in this situation.

Naveen3Singh:
Do you want it to be same as the settings even if it looks bigger here?

tibbi:
no need, it is ok like this. Looks perfectly fine for the initial implementation, we will see what feedback will it get. Thanks

PR #1944, opened 2023-01-17
Shortened no upcoming events string (#1942)

### LINKED ISSUE #1942: (Nitpick) 'Seems like you don't have any upcoming events' seems a bit verbose/redundant
Screenshot_20230116-224517_Trebuchet

Personally I'd be fine if instead of this sentence simply nothing gets displayed if there are no upcoming events. :)
Or if it need be, maybe make it at least a bit shorter, like 'No upcoming events'?

Aga-C:
Closes #1942


nofishonfriday:
Thank you for implementing it so quickly.

PR #1984, opened 2023-02-20
Update taiwan.ics for 2023 holidays

typebrook:
Ref: https://data.gov.tw/dataset/14718

New ics file is validated by https://github.com/collective/icalendar


tibbi:
should be good, thanks

Issue #360, opened 2018-02-12
support per-event colors of CalDAV calendar


MyLogins:
DAVdroid support per-event colors for events (RFC 7986 COLOR) since version 1.7, but unfortunately Simple-Calendar handle all events from a CalDAV calendar as one event type with the same color.

532910:
Not only event, but whole calendar color should be taken from CalDAV. Same RFC 7986.
```
# cat .Radicale.props 
{"tag": "VCALENDAR", "ICAL:calendar-color": "#b7d841ff", "D:displayname": "calendar", "C:supported-calendar-component-set": "VEVENT,VJOURNAL,VTODO"}#
```

stavros-k:
Hey, is there any plans for this? :)

fohfuu:
Adding to this also. I can barely read the text when the colours are so random - if you have more than a couple of calendars, no colour for the background will suit all of them. It's really inaccessible for anyone with sight issues.

azzamsa:
To have multiple color, I create separate calendar in my Nextcloud calendar. 

If this issue solved, It means I can use different color types in Simple-Calendar with only one CalDav calendar?
I never know it was a thing.

allmynamesweretakenomoreideas:
My wife is using per-event colors like crazy, she has her own color scheme for different types of events and their priority (more then 10 colors). I am trying to convice her to use our own calDAV server instead of google calendar. I already managed to import her calendar with all its colors to our calDAV server. But its a no-go for her, that simple-calendar does not show these colors nor is it editable within the event options. There seems to be no other calendar on f-droid supporting this basic feature, so simple calendar with this request is as close at it gets - so I am patient and wait...it is for a good purpose ;)

Josh-Reimer:
I use Googles calendar(their calendar not their calendar app) for all my events and tasks. Simple Calendar makes all my events the same color then and I want to be able to color code some of them. Would be a nice feature.

hackrid:
Maybe one day this becomes reality :)
That's the only feature I'm missing since years.

Would be really happy about it. Thanks for your work on Simple Calendar.

Naveen3Singh:
Hello!
I have uploaded a test APK with support for CalDAV event colors here: https://github.com/Naveen3Singh/Simple-Calendar/releases/download/untagged-2c4ac5427e11b92736df/calendar-prepaid-debug.apk
[...]

PR #2041, opened 2023-04-08
Add support for CalDAV event colors

Naveen3Singh:
This PR addresses https://github.com/SimpleMobileTools/Simple-Calendar/issues/360

### Preview:
 

This reuses pieces from the CalDAV Calendar color implementation hence the similar dialog.


app/src/main/kotlin/com/simplemobiletools/calendar/pro/dialogs/SelectEventColorDialog.kt
```diff
@@ -0,0 +1,65 @@
...
+
+        view.dialog_radio_color.setFillWithStroke(color, activity.getProperBackgroundColor())
+        view.setOnClickListener {
+            viewClicked(color)
+        }
+        radioGroup.addView(view, RadioGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT))
+    }
+
+    private fun viewClicked(color: Int) {
+        if (!wasInit)
```
tibbi, line 59: always use brackets please

app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/CalDAVHelper.kt
```diff
@@ -116,35 +115,25 @@ class CalDAVHelper(val context: Context) {
...
 
-        var sortedColors = ArrayList<Int>(colors.size())
-        (0 until colors.size()).mapTo(sortedColors) { colors[it] }
-        if (sortedColors.isNotEmpty()) {
-            sortedColors = sortedColors.distinct() as ArrayList<Int>
-        }
-
-        return sortedColors
+        return colors.toSortedMap().entries
+            .associate { (k, v) -> v to k }
```
tibbi, line 136: rename k and v, it is not clear

tibbi:
nice, seems to work well. Lets just autoconfirm the new color after it is picked on that new Event color dialog, so the user doesnt have to click OK after.

Naveen3Singh:
Should I update the old calendar color dialog in settings to use this new design?

tibbi:
sure, lets do it

Naveen3Singh:
Is "Color" okay or should I add "Calendar color" string?

Changing calendar colors when using DAVx5 and Nextcloud is messy and doesn't seem to work right (in Etar too). I checked the Google Calendar app and there are no options to change colors for third-party calendars.

tibbi:
Thanks, it looks good now, keep "color". Regarding DAVx5, for each individual account, DAVx5 allows you to change the setting for "Manage calendar colors" from (1) 'Calendar colors are reset at each sync' to (2) 'Calendar colors can be set by other apps'. You have to use the second option to be able to change colors of those calendars. Isnt that your issue?

Naveen3Singh:
Ah, it was my fault. I turned off the "Manage calendar colors" so it doesn't get reset but it gets darker and darker on every sync due to the `getDisplayColorFromColor` call. I'll fix it soon.

Issue #1983, opened 2023-02-18
Non-valid or inconsistant EXDATE + Non-unique UID in .ics export


Polux49:
Hi!

I'm having troubles when exporting my data from Simple Calendar 6.21.5 Pro to .ics files. All these problems seem to be related to the `EXDATE` property, see examples below. The first of them can easily be fixed by hand. The others seem to be more intricate. 

### 1) `EXDATE` property requires `VALUE=DATE` attribute when its value is only a date (without time)

Here is an example of .ics file obtained after exporting a category from Simple Agenda. It cannot be imported into Etar or Framagenda, which consider it as non-valid. 

```
BEGIN:VCALENDAR
PRODID:-//Simple Mobile Tools//NONSGML Event Calendar//EN
VERSION:2.0
BEGIN:VEVENT
SUMMARY:Weekly reapated event with exdate
UID:b15b11ec4c07417fbf61b23c2e2372301676707315011
X-SMT-CATEGORY-COLOR:-689152
CATEGORIES:Category for test
LAST-MODIFIED:20230218T080155Z
TRANSP:OPAQUE
DTSTART:20230301T120000Z
DTEND:20230301T130000Z
X-SMT-MISSING-YEAR:0
DTSTAMP:20230218T080345Z
STATUS:CONFIRMED
RRULE:FREQ=WEEKLY;INTERVAL=1;UNTIL=20230331;BYDAY=WE
EXDATE:20230308
END:VEVENT
END:VCALENDAR
```

Replacing the line `EXDATE:20230308` by `EXDATE;VALUE=DATE:20230308` or by `EXDATE:20230308T120000Z` solves the problem. 

### 2) Events with non-unique UID

After exporting my data to a .ics file, I have found several events sharing the same UID. Here is an unmodified example (I just removed all the other events). Note that the three first events only differ by the number of `EXDATE` properties (and the `LAST-MODIFIED` property). 

Only one of these four events is visible in Simple Calendar, probably the third one as it appears at the date 20221107. When imported into other agendas (in Framagenda for this test), an error says that only two events have been imported. These two events are visible in Framagenda at the date 20221121. 

```
BEGIN:VCALENDAR
PRODID:-//Simple Mobile Tools//NONSGML Event Calendar//EN
VERSION:2.0
BEGIN:VEVENT
SUMMARY:Vulg A116
UID:20243f4dd99d4aa78581f0cfb8eef98e1658314652034
[...]

Issue #1538, opened 2021-11-10
Synchronization issue with 400 error on recurring events


NicoHood:
Hi,
I get 400 errors when synching with davx5 and radicale server. I contacted the davx5 support and we debugged it really good:

The issue happens if you
1. Create a weekly repeating event
2. Delete a single entry of that event series
3. sync

It generates the following (sample) entry, which if invalid and therefore rejected by the radicale server:
```
BEGIN:VCALENDAR
VERSION:2.0
PRODID:DAVx5/4.0-gplay ical4j/3.1.0
BEGIN:VEVENT
DTSTAMP:20211110T085217Z
UID:2b388553-e36b-44d9-a339-5fb9816aac22
SEQUENCE:1
SUMMARY:Joggen
DTSTART;TZID=Europe/Berlin:20211104T170000
DTEND;TZID=Europe/Berlin:20211104T170000
RRULE:FREQ=WEEKLY;INTERVAL=1;BYDAY=TH
STATUS:CONFIRMED
BEGIN:VALARM
TRIGGER:-PT30M
ACTION:DISPLAY
DESCRIPTION:Joggen
END:VALARM
END:VEVENT
BEGIN:VEVENT
DTSTAMP:20211110T085217Z
UID:2b388553-e36b-44d9-a339-5fb9816aac22
RECURRENCE-ID;TZID=Europe/Berlin:20211118T170000
SEQUENCE:1
DTSTART;TZID=Europe/Berlin:19700120T004731
DTEND;TZID=Europe/Berlin:19700120T004731
EXDATE:20211117T230000Z
END:VEVENT
BEGIN:VTIMEZONE
TZID:Europe/Berlin
LAST-MODIFIED:20201010T011803Z
BEGIN:DAYLIGHT
TZNAME:CEST
TZOFFSETFROM:+0100
TZOFFSETTO:+0200
DTSTART:19160430T230000
RDATE:19400401T020000
RDATE:19430329T020000
RDATE:19460414T020000
RDATE:19470406T030000
RDATE:19480418T020000
RDATE:19490410T020000
RDATE:19800406T020000
END:DAYLIGHT
BEGIN:STANDARD
TZNAME:CET
TZOFFSETFROM:+0200
TZOFFSETTO:+0100
DTSTART:19471005T030000
RRULE:FREQ=YEARLY;UNTIL=19491002T010000Z;BYMONTH=10;BYDAY=1SU
END:STANDARD
[...]

Issue #1358, opened 2021-04-25
Editing single instances of recurring events glitchy


vocux:
Hi! I'm using v6.13.5 and syncing Calendars via CalDAV / DAVx5 / Nextcloud (and using Thunderbird as another client).
Editing single instances of recurring events seems glitchy or incompatible with my other clients... in two ways:

1. When editing a single instance from a series of repeating events from inside the app, on all my other clients (Nextcloud calendar app or Thunderbird for example) this results in two events being shown, one from the series (without any changes applied) and the new one with the changes applied. There should be only one, right?

2. When editing a single instance from within Thunderbird or the Nextcloud calendar app, all other clients reflect this change correctly, however in Simple Calendar, this changed event is now completely missing, and no events are shown that day!

Since all my other clients handle these edits of single instances as expected I figured that the problem might lie with Simple Calendar.

If you need any help reproducing this, let me know.
Best wishes!

PinkPlasticBeach:
I confirm this bug. This bug is crucial!

Because other people in shared calendar see me signing up for a shift that i wanted to sign out from. After picking the wrong day.
Me being unaware of this as simple calendar shows only the correct edited Events and not the leftover false one.


I use,too:
simple calendar 6.15.1
+ caldav 3.3.12-ose
+ nextcloud

Its reproduceable:

- Create Event on monday - repeat weekly

- Edit Event to tuesday - confirm Change for this instance and all Future instances

> In simple calendar i can only see the correctly changed instances on Tuesday.
In "etar"-app I can also see the one leftover instance on Monday I edited.

Problem seems to be the deletion of the edited isntance online.

This reproduces only with "change this instance and future ones".

If if I choose "change all instances". The edited instance will be deleted correctly but with it all earlier instances.
So that is a WORKAROUND for this bug: if you choose the first "alive" instance of the events, you will edit all future events and only delete past instances.
but it's another bug itself.


Please fix this soon. It keeps me from recommending simple calendar to friends, because it can produce real problems in real-life ;)

Thanks for the app

yannik1015:
This big still exists in version 6.20.1 as described above.

aykarpov:
Hello

I noticed this feature when editing a recurring event.

For example, I am creating an event that will repeat on Mondays. If I reschedule this event to Thursday, the recurring day of the event will also immediately change to Thursday.

[...]

Issue #1017, opened 2019-11-19
Modified occurrence of a repeating event not shown


tobiscode:
Hi all,

I use DAVx5 to synchronize my hosted CalDAV calendars to my phone. I use repeating events, where I modify single occurrences (mostly adding descriptions). This will lead to SimpleCalendar not showing the modified occurrence anymore at all. Regular repeating events that are not modified work as expected.

I made sure it's not a DAVx5 - the events appear correctly if I use the Android (Google) Calendar app (which for now, I have to use because of this bug). They also appear correctly in the Web UI of my provider and Thunderbird (where I create them).

Possibly related to #950, #735,  #981 by glancing over them.

Android 10
SimpleCalendar 6.6.5
DAVx5 2.6-ose

austinhuang0131:
Can confirm, Android 9 / v6.11.3 / v3.3.7-ose

PR #2053, opened 2023-04-20
Don't set EXDATE on recurrence exceptions

evan-goode:
Resolves (or at least improves):
- #1538
- https://github.com/bitfireAT/davx5-ose/discussions/66
- https://github.com/Kozea/Radicale/issues/1264
- May also resolve some of the issues described in #1983.

The "EXDATE" property is only meant to be set on events with the "RRULE" property. See discussion in #1538, especially the response from DAVx5 support listed in the first comment.

I tried also setting STATUS_CANCELED on these recurrence exception events, but DAVx5 would sometimes no longer recognize them as belonging to the parent event and would give the child events different UIDs, which is undesirable. It seems like it works great to just leave out the EXDATE.

I have not tested this patch with other CalDAV setups, but it solves the problem for me on DAVx5+Radicale.


tibbi:
@Naveen3Singh can you please test this?

Naveen3Singh:
It works. Repeating exceptions are no longer rejected by Radicale and Mailbox. Nextcloud works just as before.

> I tried also setting STATUS_CANCELED on these recurrence exception events, but DAVx5 would sometimes no longer recognize them as belonging to the parent event and would give the child events different UIDs, which is undesirable

I couldn't reproduce the UID issue when setting the status to STATUS_CANCELLED but as much as I could test, deletion seems to work fine without the canceled event status.

There are still other issues related to recurring event exceptions. For example, when a recurring event occurrence is modified (not deleted) by some other calendar app, the events disappear in Simple Calendar as mentioned in https://github.com/SimpleMobileTools/Simple-Calendar/issues/1017
Other related/similar issues: https://github.com/SimpleMobileTools/Simple-Calendar/issues/981 (reproducible with local events too), https://github.com/SimpleMobileTools/Simple-Calendar/issues/950, https://github.com/SimpleMobileTools/Simple-Calendar/issues/735, https://github.com/SimpleMobileTools/Simple-Calendar/issues/1358

tibbi:
alright, thanks. Maybe @Naveen3Singh you could take at those related repeating issues, will create a task for them

Naveen3Singh:
> alright, thanks. Maybe @Naveen3Singh you could take at those related repeating issues, will create a task for them

Sure :)

Issue #2051, opened 2023-04-17
[QUESTION] What event triggers data backup in latest version?


HarriBuh:
While I'm very happy to see this feature finally getting to Simple Calender, there is no hint or option how to specify the backup action itself yet. Since the update the app hasn't done a single backup, for example.

I ask for a more detailed explanation and option inside the app to determine it, like a schedule (max. amount of backups per day etc.)
The current function is almost useless, because it doesn't work at all.

tibbi:
it triggers at 6 AM every day (or at the time the device is turned on), there are no other triggers yet. We just implemented the easiest solution to begin with.

HarriBuh:
The app does not create an automated update, either when I turn on my phone (~ 6 am every morning), nor at a later hour.

HarriBuh:
Android 13
App version 6.21.7
I do use Android's energy saving, BUT this doesn't affect other backup app doing their jobs effectively.
So should I try disabling battery saving for your app and observe it's behaviour again?

P.S.: And you are 10% certain this issue is not related to Android 13‘s quite aggressive rights management?

tibbi:
it is hard to say what is breaking such features, there are just too many possible device or system specific issues :/ Maybe check https://dontkillmyapp.com/ for some potential issues.

HarriBuh:
I just rebooted my phone and got a tiny error notification from your app. It was too fast to read, but it definitely contained something with "java" and "Io".

tibbi:
does manual exporting work from the top Menu -> Export events?

HarriBuh:
Yes, I have been doing out this way for months, no issues.

Naveen3Singh:
@HarriBuh Hi!

Can you share a screenshot of your "Manage Automatic Backups" dialog configuration? 
(The IO error might happen if the export file already exists and it was created by some other app or an uninstalled version of Simple Calendar)

HarriBuh:
> @HarriBuh Hi!
> 
> Can you share a screenshot of your "Manage Automatic Backups" dialog configuration? (The IO error might happen if the export file already exists and it was created by some other app or an uninstalled version of Simple Calendar)

Sure:

![markup_1000008560](https://user-images.githubusercontent.com/123186113/233800498-5b172ad2-e291-4e5c-a4d9-cd2d4e59d7c1.png)

Naveen3Singh:
[...]

PR #2056, opened 2023-04-23
Fix file permission issue related to automatic backup feature

Naveen3Singh:
Closes https://github.com/SimpleMobileTools/Simple-Calendar/issues/2051

The exports will still fail if the user manually deletes the provided backup folder, currently an error toast is shown but this specific error can be handled better using a notification.


tibbi:
thanks, hope it will work well :) Removed some extra spaces at switch/case, they arent needed there

PR #2059, opened 2023-04-25
Use `adjustResize` flag

Naveen3Singh:
This fixes issues with obscured views behind the keyboard in EventActivity.

See https://github.com/SimpleMobileTools/Simple-Commons/pull/1667#issuecomment-1527471639 for details.


app/src/main/AndroidManifest.xml
```diff
@@ -132,6 +132,7 @@
             android:name=".activities.EventActivity"
             android:exported="true"
             android:label="@string/new_event"
+            android:windowSoftInputMode="adjustResize"
```
tibbi, line 135: is this really needed? Updating commons seems to be enough to me as it fixed both EventActivity and TaskActivity

Naveen3Singh: Without `adjustResize`, if I try to edit attendee's field, it looks like this:

The soft input mode defaults to `adjustPan` which hides the toolbar and unnecessarily adds some extra padding at the bottom.

With `adjustResize`, everything looks fine except that the attendee field doesn't scroll up automatically (so it's above the keyboard), one has to do it manually.

Naveen3Singh: Resolved the conflicts, if the screenshot above looks okay (not to me) then this PR can be discarded.

tibbi:
will take a look at it again, but whatever flag you add, it should be added both to event and task activity just to avoid some future glitches

tibbi:
works fine to me without the flag too, guess it will be some device specific glitch. But if it fixes something to you and doesnt break anything for me, guess we should have it

Issue #1698, opened 2022-04-22
Changed recurrence events are not shown


ccplatz:
Hi guys,
when I change the starting time in Nextcloud or Etar of a recurrence event it completly disappears from simple calendar. In NC or Etar it is still there, of course in the changed version. When I change it in SC it works as expected.

I use version 6.18.1 Pro.

Thanks for help.

tibbi:
can you reproduce it reliably?

ccplatz:
Yep, I tried it with different apps on different series. They always disappear in SC and not in the others. I also deleted all data.

tibbi:
please tell exact steps for reproducing, just to ensure we get it right

ccplatz:
* create an recurring event in any client app
* change the starting time of one item in any client app, but not in SC
* reload the calendars in SC
* the changed item disappears

aurisnoctis:
Thanks, it is fixed for me with Simple Calendar version `6.22.0 Pro` available on F-Droid.

All missing changed single items of calendar series magically reappeared now 🙌. I am using DAVx5 and Nextcloud (as already tested by the provider of the bug fix pull request).

Thank you very much for the calendar and the bug fix. 👍

PR #2086, opened 2023-05-22
Fixed recurring event repetition exception issue

nelath:
Resolves:

- https://github.com/SimpleMobileTools/Simple-Calendar/issues/1698

When a repetition exception is added to the event which itself is an exception to the repetition, it is not shown in the calendar.

Successfully tested this with DAVx5 and Nextcloud.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/CalDAVHelper.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/CalDAVHelper.kt
@@ -264,9 +264,6 @@ class CalDAVHelper(val context: Context) {
                     parentEvent.addRepetitionException(originalDayCode)
                     eventsHelper.insertEvent(parentEvent, addToCalDAV = false, showToasts = false)
 
-                    event.parentId = parentEvent.id!!
-                    event.addRepetitionException(originalDayCode)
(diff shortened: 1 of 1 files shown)


tibbi:
you should test it with some Google calendar too

nelath:
Now also successfully tested with Google calendar

Naveen3Singh:
Other than the comment I left, this solution works fine. Tested with Nextcloud, Mailbox, and Google Calendar.

This happens because of the way we handle deleted recurrences at the presentation level. An exception event with [exceptions is treated as a deleted occurrence](https://github.com/SimpleMobileTools/Simple-Calendar/blob/dea55ca94c7a183ae0fd14c333a42b121443dbf4/app/src/main/kotlin/com/simplemobiletools/calendar/pro/helpers/EventsHelper.kt#L312) of a recurring event and thus the whole disappearing act.

This also closes https://github.com/SimpleMobileTools/Simple-Calendar/issues/1017

Naveen3Singh:
In rare cases, this can help resolve reports about zombie notifications. Users may mistakenly believe they have deleted all events, unaware that exceptions of recurring events may still exist in the database, potentially triggering notifications.

tibbi:
okay thanks

Naveen3Singh:
I'll apply my suggested changes in another PR.

Issue #981, opened 2019-09-22
Editing repeating event *must not* delete past events of the series


xss:
Say we have a repeating event that started on September 1. Now, when editing any later event of this series, let's say today's occurrence, and upon saving selecting "Update all occurrences", Simple Calendar will update all future occurrences of the series, but will delete all this series' occurrences *before* the selected date -- today in my example.

In my opinion, this is *not* the expected behavior of an option labelled "Update all occurrences". All occurrences should be updated equally, independent of whether they are future or past occurrences in the same series. As Simple Calendar doesn't have an option like "Update only future occurrences", this comes as even more of a surprise.

I didn't see anything "suspicious" concerning this issue in the corresponding dialog's code file. I suspect, "Update all occurrences" sets a new start date for the series --hence doesn't explicitly *delete* before the selected date--, which causes the previous occurrences to be lost.

Due to another long-standing bug in Simple Calendar --all reminders stop working unrecoverably after every time Android's battery saver feature got turned on; the only fix being to re-edit all events in Simple Calendar-- I'm forced to use this "Update all occurrences" functionality rather regularly.

Simple Calendar v6.6.1
Android 6.0.1
Event is on a CalDAV synced NextCloud calendar, not a "local" one

nic3-14159:
> I didn't see anything "suspicious" concerning this issue in the corresponding dialog's code file. I suspect, "Update all occurrences" sets a new start date for the series --hence doesn't explicitly _delete_ before the selected date--, which causes the previous occurrences to be lost.

I've looked at the code, and this does indeed seem to be the case. I believe I have a fix for this, I just need to test it and make sure it doesn't break anything else, and then I'll open a pull request.

Aga-C:
I have the same issue with local calendar.

dumblob:
@nic3-14159 did you manage to take a look at the code and maybe find out how to patch it?

Naveen3Singh:
Hello!

If anyone is up for testing, here's a test APK with a fix: https://github.com/Naveen3Singh/Simple-Calendar/releases/download/fix_981/calendar-prepaid-debug.apk

Please let me know if it works for you or if it causes any new issues.

 @xss  @nic3-14159  @Aga-C @dumblob @Ordoviz @itaton

Aga-C:
@Naveen3Singh It works for me with local calendar 👍

PR #2093, opened 2023-05-28
Properly handle `Update all occurrences` option

Naveen3Singh:
Closes https://github.com/SimpleMobileTools/Simple-Calendar/issues/981

The events were moved forward because the start and end timestamps from the repeating occurrence were saved instead of the original start and end times.

The call to `eventsHelper.addEventRepeatLimit()` was removed because I believe it was added there by mistake and the changes were overwritten anyways due to the next call to `eventsHelper.updateEvent()`. Another reason to remove it is to avoid triggering a CalDAV update which may interfere with the event update.

### Video

https://github.com/SimpleMobileTools/Simple-Calendar/assets/36371707/8ec47161-2639-4c08-9163-c3f15a145b51


tibbi:
okay thanks, lets hope it wont create any unexpected issues

PR #2126, opened 2023-06-24
Properly handle `Update/Delete this and all future occurrences`

Naveen3Singh:
This PR fixes the following issues related to editing recurring events:

**Bug 1:**
 - Create a recurring event with a specific number of occurrences like 10 that starts on Jan 1
 - Edit some occurrence e.g. 5th and select "Update this and future occurrences only"
 - The edited event (5th) kept repeating 10 times instead of ending on the end date of the original event (10th Jan). This was not consistent with other calendars (except Nextcloud, seems like it's a bug there too).

**Bug 2:**
When choosing `Update/Delete this and all future occurrences`, future occurrences were deleted/updated but the selected occurrence was never deleted/updated. Only reproducible with all-day events.

**Bug 3:**
When choosing `Update this and all future occurrences`, everything was updated properly but the event was duplicated on the date of the selection. Only reproducible with all-day events.

**Bug 4:**
When creating a recurring event with a date as a repeat limit, the event ends a day before the date it actually should. This is only reproducible with non-all-day CalDAV events.

_Regarding the `UNTIL` DateTime timezone, [this SO answer](https://stackoverflow.com/questions/34913427/icalendar-until-rule-must-be-of-the-same-type-as-dtstart-property) and the format used by Google calendar was taken into consideration._

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/EventActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/activities/EventActivity.kt
@@ -1382,6 +1382,10 @@ class EventActivity : SimpleActivity() {
                         mEvent.id = null
                         eventsHelper.apply {
                             addEventRepeatLimit(eventId, mEventOccurrenceTS)
+                            if (mEventOccurrenceTS == originalEvent.startTS) {
+                                deleteEvent(eventId, true)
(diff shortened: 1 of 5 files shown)


tibbi:
did a couple basic tests and nothing too obvious broke, hope it will be good :) Thanks

Issue #582, opened 2018-08-13
Set any day as first of the week


bb1a2:
Not just Sun/Mon

Thanks

fuentesj11:
Is this something you would really like to add to the app? I went through the code and you have a Boolean controlling either Monday or Sunday being first day of the week. This goes back to the BaseConfig class from your imported package/project commons.helper. The point being the project has little room for extensibility for this feature to be implemented without modifying the imported project and its associated classes and logic.

tibbi:
most features require a lot of work, not just changing a line :)

fuentesj11:
I think you misunderstood what I was trying to say. To my knowledge there is no access to the commons module. The commons modules needs to be modified (which I don't have access to) to implement any type of change. The alternative is to modify what is available in the package that is "open" but that changes your logic and separation of tasks that you currently have implemented.

tibbi:
it doesnt need to be modified. I can just create a new field in the app Config and use that.

fuentesj11:
Yes but now you have a field that is not used in BaseConfig which brings the question as to why is the isSundayFirst in BaseConfig instead of the app Config if the field is the app. I would assume the functionality of isSundayFirst is important in more than one app and thus the functionality of starting on any day besides Monday or Sunday may carry over as well.

fuentesj11:
Sorry I was just thinking since isSundayFirst is in the BaseConfig with is in a commons modules that any new functionality like starting on any day of the week might be shared, like starting on Monday or Sunday, by multiple projects that use commons module. I will just stick with a field in Config though.

hubertbanas:
Being able to set any day as first of the week would be very helpful.

naveensingh:
Hey there!

If anyone is interested in testing out this new feature, I've made it available as a debug APK, which you can download from this link: [APK Link](https://github.com/naveensingh/Simple-Calendar/releases/download/any_day_week/calendar-core-debug.apk)

If you prefer to compile the APK yourself, you can find the corresponding pull request here: [Pull Request Link](https://github.com/SimpleMobileTools/Simple-Calendar/pull/2159)

Feel free to provide feedback if you encounter any issues. Your input is much appreciated!

Thanks.

PR #2129, opened 2023-06-29
Rename method `getDatesWeekDateTime()` to `getFirstDayOfWeek()`

Naveen3Singh:


tibbi:
so we are ready for using any day of the week as the first day? For https://github.com/SimpleMobileTools/Simple-Calendar/issues/582 :) It would need a new setting instead of the current toggle.

WessellUrdata:
> so we are ready for using any day of the week as the first day? For #582 :) It would need a new setting instead of the current toggle.

It's doable, just keep the Joda-time's Monday as first day behavior in mind ☺️
One problem that I can think of is the calculation of the week number, but that shouldn't be too hard to deal with

tibbi:
ok :)

Naveen3Singh:
agreed, shouldn't be too hard to do.

PR #2132, opened 2023-06-30
Changed event type and color order (#2131)

### LINKED ISSUE #2131: Unintuitive order of event color and event type settings
When creating a new event, you have the following screen:

I feel the order of color and event type is not intuitive and should be switched. So type selection is above.

Reason for this is, I defined color with the event type. I have no need to manually override this color, but the event type is changed quite often and I frequently mistap.

Aga-C:
Closes #2131


tibbi:
looks good, thanks

PR #2133, opened 2023-06-30
Ensure event type visibility when a new event or task is created

Naveen3Singh:
We are doing it in `insertEvent()`, `insertTask()` methods because it's cleaner than doing it in `EventActivity` and `TaskActivity`.

The event types stay disabled if the user adds new events using some other client, the event types are enabled only if the event is added using Simple Calendar.


tibbi:
looks good, thanks

PR #2145, opened 2023-07-09
- Improve UI Monhtly and daily view for landscape

edgy-b:
General UI usability improvement for landscape in Monthly and daily view

**Before**
before

**After**
after


tibbi:
looks nice, thanks

PR #2152, opened 2023-07-15
Added editing options for repeating events in weekly view (#2122)

### LINKED ISSUE #2070: An event occasionally gets invisible if dragged back into the same place
After an app restart, it's displayed well again.
https://user-images.githubusercontent.com/2134486/236712641-fbb570ac-6cb0-48e3-b3bd-2386a91c8364.mp4

### LINKED ISSUE #2122: Drag-n-Drop repeated events moves the first event
#### Steps to reproduce

1. Switch to weekly view
2. Create a new repeated event (e.g. starting at 11:00), recurring every day
3. Move the second instance of it to one hour later (e.g. 12:00)

#### Expected

- The 1st instance should stay where it is, as it was not touched (e.g. 11:00)
- The 2nd instance should be moved to the selected time (e.g. 12:00)
- The 3rd and further instances should stay where they are (e.g. 11:00)

#### Actual

- All instances were moved to the selected time (e.g. 12:00)
- The original 1st instance seems removed. Now the repeated event begins when the 2nd instance was before.

#### Screenshots & Recording

<details>
<summary>Before & After Screenshots</summary>

##### before
Screenshot_20230619-104659_Calendar

##### after
Screenshot_20230619-104824_Calendar

</details>

<details>
<summary>Recording</summary>

https://github.com/SimpleMobileTools/Simple-Calendar/assets/1645646/1885eb47-a447-40f4-94e5-208be07f8b3e

</details>

Aga-C:
Closes #2122

- Added `EditRepeatingEventDialog` to moving events and tasks in weekly view.
- Moved code for editing recurring events and tasks to `EventsHelper`to reuse it in `EventActivity`, `TaskActivity` and `WeekFragment`.
- Changed refreshing method in `WeekFragment` for case of editing all or future occurrences - `updateCalendar()` was refreshing only current fragment and there were another cached in `MainActivity` that still had old data, so in this case we have to refresh all fragments.
- These changes caused a small bug in weekly view - when user closed the `EditRepeatingEventDialog` without selecting anything, the event has disappeared. To fix this, I've added another case in `EditRepeatingEventDialog` to return `null` in callback, when the user hasn't selected anything. It doesn't affect `EventActivity` and `TaskActivity`.
  - ~~Also, to make this work, I had to force events' refreshing in `WeekFragment`. I'm doing this by resetting `lastHash` value.~~ EDIT: Instead of force refresh, I'm reverting dragged view. I've also added it to the case when event hasn't been changed, so it also fixes #2070

https://github.com/SimpleMobileTools/Simple-Calendar/assets/85929121/02c1b1c2-7d78-42f8-8eb1-aee5a774ef51

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt
@@ -191,6 +191,7 @@ class WeekFragment : Fragment(), WeeklyCalendar {
 
     fun updateCalendar() {
         if (context != null) {
+            currentlyDraggedView = null
             WeeklyCalendarImpl(this, requireContext()).updateWeeklyCalendar(weekTimestamp)
         }
     }
@@ -299,9 +300,7 @@ class WeekFragment : Fragment(), WeeklyCalendar {
                                                     activity.hideKeyboard()
                                                     when (it) {
                                                         null -> {
-                                                            // force update by removing hash
-                                                            lastHash = 0
-                                                            updateCalendar()
+                                                            revertDraggedEvent()
                                                         }
                                                         EDIT_SELECTED_OCCURRENCE -> {
                                                             context?.eventsHelper?.editSelectedOccurrence(newEvent, false) {
@@ -323,8 +322,12 @@ class WeekFragment : Fragment(), WeeklyCalendar {
                                                 }
                                             }
                                         } else {
-                                            context?.eventsHelper?.updateEvent(newEvent, updateAtCalDAV = true, showToasts = false) {
-                                                updateCalendar()
+                                            if (event.startTS == newEvent.startTS && event.endTS == newEvent.endTS) {
+                                                revertDraggedEvent()
+                                            } else {
+                                                context?.eventsHelper?.updateEvent(newEvent, updateAtCalDAV = true, showToasts = false) {
+                                                    updateCalendar()
+                                                }
                                             }
                                         }
                                     }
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt
```diff
@@ -284,14 +287,45 @@ class WeekFragment : Fragment(), WeeklyCalendar {
...
+                                        )
+                                        if (event.repeatInterval > 0) {
+                                            val activity = this.activity as SimpleActivity
+                                            activity.runOnUiThread {
+                                                EditRepeatingEventDialog(activity) {
+                                                    activity.hideKeyboard()
+                                                    when (it) {
+                                                        null -> {
+                                                            // force update by removing hash
+                                                            lastHash = 0
```
Naveen3Singh, line 303: As we are doing this in the background, the event disappears because we already returned `true` in `setOnDragListener` which caused the view to be hidden by this code:
https://github.com/SimpleMobileTools/Simple-Calendar/blob/b2236a2a03f3772290bf26d91d3a15d25dd53c53/app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt#L897-L903

If we address this concurrency issue properly or if we choose to always force refresh when dragging events (repeating and normal) while doing drag-and-drop, #2070 will be resolved.

Aga-C: I've fixed it by not nulling `currentlyDraggedView` here, but in the action itself. Can you check if it's ok?

tibbi:
ok, thanks

app/src/main/kotlin/com/simplemobiletools/calendar/pro/fragments/WeekFragment.kt
Naveen3Singh: Nice! It's fixed now.

Issue #2143, opened 2023-07-08
Date -1 issue when adding event in weekly view


ttes0:
After July update, having -1 day issue when adding event in weekly view. 
e.g. Selecting add event on 2023-07-08 opens in edit event screen with date set to 2023-07-07

(on android v.11)

tibbi:
do your weeks start with sunday or monday? Does it matter if you create the event at 2:00 or 22:00?

ttes0:
> do your weeks start with sunday or monday? Does it matter if you create the event at 2:00 or 22:00?

Monday, but I've tried setting either. I've tried setting to 24hr and 12hr clock. I'm unable to downgrade to May version of apk, but I've uninstalled and reinstalled the July version of apk.
I have apk installed on 2 phones, and same issue.

Naveen3Singh:
what timezone are you in?

ttes0:
GMT-7, but when selecting inside app, only shows GMT-8

ttes0:
Selected is 'America/Los_Angeles'

ttes0:
> > do your weeks start with sunday or monday? Does it matter if you create the event at 2:00 or 22:00?
> 
> Monday, but I've tried setting either. I've tried setting to 24hr and 12hr clock. I'm unable to downgrade to May version of apk, but I've uninstalled and reinstalled the July version of apk. I have apk installed on 2 phones, and same issue.

Downloaded May version from github, but get 'app not installed' when trying to install.

Naveen3Singh:
@WessellUrdata this seems related to your recent fix https://github.com/SimpleMobileTools/Simple-Calendar/pull/2077

WessellUrdata:
> @WessellUrdata this seems related to your recent fix #2077

looking into it.

Naveen3Singh:
> I am able to reproduce the bug in the latest version of the app 6.22.1. Cannot be reprduced in 6.21.9. I probably broke it. my bad, trying to fix it now

hey, I'll be working on this so no worries if you can't fix it. Also, everyone breaks things all the time. These kinda of things often have a domino effect.

WessellUrdata:
> > I am able to reproduce the bug in the latest version of the app 6.22.1. Cannot be reprduced in 6.21.9. I probably broke it. my bad, trying to fix it now
> 
> hey, I'll be working on this so no worries if you can't fix it. Also, everyone breaks things all the time. These kinda of things often have a domino effect.

[...]

Issue #1894, opened 2022-12-04
Weekly view doesn't display correct week when setting "Start week on Sunday" enabled


maze88:
Recreation:
The bug occurred for me today (_Sunday_, 2022-12-4), where my calendar's weekly view was still on the previous week (2022-11-27 through 2022-12-3).

Version:
6.20.3 Pro

Notes:
- Not resolved by force-stopping, and restarting the app.

tibbi:
you mean not even after restarting the app?

maze88:
> you mean not even after restarting the app?

Nope.
(I updated this detail in original post)

maze88:
Update:
At some point in the afternoon (I noticed at about 17:30) that the issue no longer existed... you may need deeper investigation on how to recreate...

tibbi:
well, guess it was related to being a Sunday yesterday, while it is a Monday already

maze88:
But yesterday [afternoon] was still totally Sunday.
I wouldn't even suspect a timezone issue, since being in +0200 wouldn't explain the issue either.

maze88:
Re-confirming this issue still occurs (I am now on version 6.21.3 Pro).

WessellUrdata:
Unfortunately, it is _somewhat_ related to timezone.

I tested several timezones by changing the time and timezone on my phone and then killing the app and relaunching it/clicking the Gtt button (Go to today) on the top bar to get the displayed week.

App version: v6.21.4 F-Droid version

The results are as follows:

**UTC+8 HKT, Hong Kong, 2023-01-29: (does NOT observe DST)**

00:00 : Gtt points to wk4 (incorrect)
20:59 : Gtt points to wk4 (incorrect)
21:00 : Gtt points to wk5 (correct)
23:59 : Gtt points to wk5 (correct)

20:59/21:00 is the cutoff.

---

**UTC+2 IST, Israel, 2023-01-29: (observes DST, but currently not in effect)**


14:59 : Gtt points to wk4 (incorrect)
15:00 : Gtt points to wk5 (correct)

14:59/15:00 is the cutoff.

---

**UTC/GMT, London, 2023-01-29: (observes DST, but currently not in effect)**

12:59 : Gtt points to wk4 (incorrect)
13:00 : Gtt points to wk5 (correct)


12:59/13:00 is the cutoff.

---

**UTC-5 EST, New York, 2023-01-29: (observes DST, but currently not in effect)**
[...]

Issue #1281, opened 2021-02-18
Fix wrong new event date in far time zones


tibbi:
more info at email from DK from New Zaeland on February 18

Issue #1213, opened 2020-10-06
Nov. 2, 2020 entries appear on Nov. 3, 2020


jwilliams0:
When I try to add a calendar entry to Nov. 2, 2020, the event appears on Nov. 3, 2020.

tibbi:
at what time are you creating it? What app version are you using?

jwilliams0:
Version 6.10.3.  I have tried many different times and durations, and they all appear on Nov. 3, instead of Nov. 2.

Thanks

Oct. 7, 2020 2:54:56 a.m. Tibor Kaputa <notifications@github.com>:

tibbi:
no other date has this issue? How about repeating events?

jwilliams0:
I haven't noticed this with any other date.  Yes, even repeating events schedule for Nov. 2, appear on Nov. 3.  Are you able to display events on Nov 2?

Oct. 7, 2020 9:19:44 a.m. Tibor Kaputa <notifications@github.com>:

tibbi:
ye sure, and not a single person reported any issue like that.

shanks1337:
I have the same Issue and I've screen recorded it. See https://imgur.com/a/l8aadFC

v 6.10.3 Pro from F-Droid. 
German Language Setting
Android 11

Video 1: Note that I clicked on _June 5th_ and when I try to add a new appointment it is _June 6th_. I restarted my phone and app and it seemed persistent. But now it seem to be fine for now.

Video 2: The event goes clearly over 2 days and is not properly displayed on the month view. It didn't work for other dates as well when I created new events. Now seems fine again just like the one in video 1.  I can still reproduce it for this event as you can tell from the video. If I deleted it and re added it, my guess would be that it would probably work properly.

tibbi:
in what views does it happen. Only the monthly?

jwilliams0:
It happens in the week view.  The entry appears on the correct day in the month and day views.

Oct. 25, 2020 5:40:10 p.m. Tibor Kaputa <notifications@github.com>:

tibbi:
I see, do you have 7 days shown at the weekly view, or some other amount? Do you have week number displayed enabled in the app settings?

jwilliams0:
I have 7 days showing, and yes, the week number is displayed.

Thanks,
Joy

Oct. 25, 2020 5:52:51 p.m. Tibor Kaputa <notifications@github.com>:

algorithminflux:
[...]

PR #2154, opened 2023-07-15
Use local timezone to configure weekly views 

Naveen3Singh:
Closes https://github.com/SimpleMobileTools/Simple-Calendar/issues/2143

##### Why did this happen?
<details> 
<summary>
Click to expand

</summary>

1. The `MainActivity` retrieves the start-of-week in UTC using the `getFirstDayOfWeek()` method, resulting in a timestamp like `2023-07-16T00:00:00.000Z`.
2. The start-of-week timestamp is then passed to the `WeekFragmentsHolder` as an argument called `WEEK_START_DATE_TIME`.
3. In the `WeekFragmentsHolder`, the `WEEK_START_DATE_TIME` is parsed and stored as `currentWeekTS` in seconds. The `DateTime.parse()` method handles time zones during parsing.
4. The `getWeekTimestamps(currentWeekTS)` method is used to generate start-of-week timestamps for all prefilled weeks and is passed to the `MyWeekPagerAdapter`.
5. The UTC start-of-week timestamps are then passed down to the `WeekFragment` using the `WEEK_START_TIMESTAMP` argument.
6. In the `WeekFragment`, when creating a new event, the event start timestamp is calculated based on the start-of-week timestamp, and the `EventActivity` is launched with the `NEW_EVENT_START_TS` argument.
7. In the `EventActivity`, the `NEW_EVENT_START_TS` is treated as if it's in the local timezone, causing the created event to be shifted forward or backward by a day due to potential time zone differences.

</details>

##### Why not handle everything in UTC instead?

The weekly view originally used the local timezone to configure everything but was modified to use UTC only to fix a `Start week with Sunday` issue https://github.com/SimpleMobileTools/Simple-Calendar/issues/1213 where events were created on the wrong day. Since `getFirstDayOfWeek()` was rewritten to handle week preferences properly, there is no need to complicate things using UTC anymore (otherwise we'll have to handle UTC-to-local-timezone conversion in many cases).

This also fixes an issue with the drag-and-drop feature where the events were moved by ±1 day because the difference in timezone was not taken care of.

I have tested this with many time zones and made sure the related issues (#1894, #1281, #1213) don't pop back. I also carefully examined all usages of `getFirstDayOfWeek()` and did not find anything that could break by switching to local timezone.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/extensions/Context.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/extensions/Context.kt
@@ -758,6 +758,10 @@ fun Context.editEvent(event: ListEvent) {
 }
 
 fun Context.getFirstDayOfWeek(date: DateTime): String {
+    return getFirstDayOfWeekDt(date).toString()
+}
+
+fun Context.getFirstDayOfWeekDt(date: DateTime): DateTime {
     var startOfWeek = date.withTimeAtStartOfDay()
     if (!config.startWeekWithCurrentDay) {
         startOfWeek = if (config.isSundayFirst) {
@@ -771,7 +775,7 @@ fun Context.getFirstDayOfWeek(date: DateTime): String {
             startOfWeek.withDayOfWeek(DateTimeConstants.MONDAY)
         }
     }
-    return startOfWeek.toString()
+    return startOfWeek
 }
 
 fun Context.isTaskCompleted(event: Event): Boolean {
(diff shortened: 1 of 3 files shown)


tibbi:
so you are still working on it?

Naveen3Singh:
I'm done now, I found that the drop-n-drop always drops an event at **hour:minute** instead of respecting the drop hour location and altering the start minute but that's more like a feature-not-implemented thing:

https://github.com/SimpleMobileTools/Simple-Calendar/assets/36371707/e0a1abca-5402-4174-99cc-6ad19b24cfb1

Naveen3Singh:
There's also https://github.com/SimpleMobileTools/Simple-Calendar/issues/2070, not sure why it's happening (the `ACTION_DROP` case is triggered but it shouldn't).

tibbi:
those arent that major problems now, events shifted by 1 day are much worse

tibbi:
this should be enough for updating the app later on today, thanks

Naveen3Singh:
https://github.com/SimpleMobileTools/Simple-Calendar/pull/2152 could be added to the release too, the changes there don't look too dangerous.

tibbi:
ok thanks

PR #2159, opened 2023-07-22
Allow starting week with any day

naveensingh:
Closes https://github.com/SimpleMobileTools/Simple-Calendar/issues/582

### Changes and notes:
 - Added "Start week on" preference. This replaces the previous "Start week on Sunday" preference. By default, the locale's default first day of week is used (the previous "Start week on Sunday" configuration is ignored). In the weekly view, "Start week with current day" takes precedence over this preference.
 - Updated `Context.getFirstDayOfWeek()` logic again to handle any day as the first day of the week (affects the Weekly view).
 - Updated the way `MonthlyFragment`, and `YearFragment` calculate the month column index for the first day of the month (to handle any day as the first day of the week).
 - Update the `isWeekend()` logic used to highlight weekends when "Highlight weekend on some views" is enabled. Saturday and Sunday are still highlighted as weekends no matter what day is selected in the "Start week on" preference.
 - Updated the Weekday sorting logic used when configuring repeat rules in event and task activities so that the weekdays are shown in proper order when using any day as the first weekday. The `getSelectedDaysString()` and `moveLastItemToFront()` methods will stay in **Simple-Commons** as they are used by **Simple-Clock**.
 - `BaseConfig.isSundayFirst` will also remain in **Simple-Commons** as it's still used by **Simple-Clock**.


naveensingh:
Tested all changes both manually and by using test code wherever possible (e.g. programmatically tested `Context.getFirstDayOfWeek()` with all hours, timezones, and weekdays). Also tested known weekly view cases like #1894

naveensingh:
seeing some timezone issues in the weekly view, not yet sure if it's caused by the new code but let me check it before this gets merged

naveensingh:
It was nothing, please proceed.

tibbi:
this looks actually good, havent found any issues. Thanks

PR #2171, opened 2023-08-11
Localized date format

### LINKED ISSUE #2127: Bug: date formats only match locale settings in US English
This is NOT a duplicate of https://github.com/SimpleMobileTools/Simple-Calendar/issues/1055 - that issue refers to a feature request for overriding locale settings to show date formats according to user-defined preferences (@tibbi confirms in the thread that date format customization is indeed on the roadmap.)

This bug report does not refer to customisation: it concerns respecting the user's locale settings. 

Currently, Simple Calendar erroneously shows US date formats (MM/DD/YYYY, September 25, 2013, etc.), irrespective of what locale is chosen: British English, German, French will all incorrectly use the US date format.

This results in major errors: to refer to the 25th day of the 9th month, Simple Calendar will in German render "Sept 25", which in a German-language contexts refers to the month of September in the year of (20)25. I don't speak every tongue under the sun, but I presume this could also be a significant issue in many other languages too.

While awaiting implementation of the locale override/date-format customisation feature, Simple Calendar must urgently implement date formats that match the user's (system-wide or app-only) locale.

hurzl:
A simplification that fixes #2127

The `context` args can be removed, too ...

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/calendar/pro/extensions/Context.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/calendar/pro/extensions/Context.kt
@@ -343,7 +343,7 @@ fun Context.notifyEvent(originalEvent: Event) {
     val displayedStartDate = when (startDate) {
         LocalDate.now() -> ""
         LocalDate.now().plusDays(1) -> getString(R.string.tomorrow)
-        else -> "${Formatter.getDateFromCode(this, Formatter.getDayCodeFromTS(event.startTS))},"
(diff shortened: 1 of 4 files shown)


hurzl:
So sorry for the semicolons. I'm sure you have the correct solution.

meskobalazs:
@tibbi Okay, maybe it was the wrong solution, but could you elaborate? Why does it look wrong? How could he fix it? Closing it with this comment seems simply rude.

PR #2186, opened 2023-08-30
changed wording

firefoxlover:
Remind until manually closed would be the translation


tibbi:
thanks, but the current version is better

firefoxlover:
hm, I agree from the programmer perspective, but I think its not clear that this is "looping". But of course your app your choice

PR #2209, opened 2023-09-25
Move migrations into app settings

esensar:


tibbi:
looks good, lets also remember the last used value at "Ignore event types in the file..." checkbox at importing

PR #2215, opened 2023-09-28
Attendee related improvements

naveensingh:
### Changes:
 - Fix attendee list visibility in event activity (I broke it during view-binding migration)
 - Replace manual attendees serialization/deserialization everywhere with a type converter so it's deserialized automatically when loading events from the database (tested upgrading from 6.23.0 in release mode, no crashes).
 - Only allow attendee status modification from the primary owner account. Previously any calendar account could update the attendee status of any other calendar account as long as they have the same account name.
 - Strikethrough events when attendee status is set to "Not going".
 - Some minor code cleanup


tibbi:
looks good, thanks

PR #2257, opened 2023-11-07
Fix escape sequence in string resource

KBenMessaoud:
I've encountered a build failure caused by an incorrect escape sequence in a string resource within strings.xml. The problematic line is: 

<string name="holidays_imported_successfully">Les jours fériés ont été correctement importés dans le type d\'évènement \'Jours fériés\'</string>

The single quote after d in d\'évènement is escaped improperly.

Solution
This Pull Request corrects the escape sequence by adding an extra backslash, ensuring that the single quote is correctly escaped. This change resolves the build failure. The corrected line is:

<string name="holidays_imported_successfully">Les jours fériés ont été correctement importés dans le type d\'évènement \'Jours fériés\'</string>

Impact
This change will allow the application to build correctly and should not impact any functionality or other parts of the application.

Please review this Pull Request and if it meets your approval, I'd appreciate it if you could merge it into the main codebase.

Thank you for considering this fix.


Aga-C:
It's a duplicate of https://github.com/SimpleMobileTools/Simple-Calendar/pull/2254.

PR #2273, opened 2023-11-26
Highlight weekends in printed matter too

ikazuhiro:
It can highlight weekend days in printed matter on weekly , monthly and yearly views.


inson1:
@ikazuhiro SMT apps were sold. Fork is here https://github.com/FossifyOrg Please create PR there

PR #2276, opened 2023-11-28
Fix for: Clicking on a day in the monthly view and selecting the wrong day

KBenMessaoud:
By altering the childHeight value to be half of the child view's measured height in our layout method, we've halved the size of the clickable area in our calendar. This change has impacted how child views are arranged, particularly the layout() method that positions and sizes views within their container. Consequently, only clicks in the lower half of the calendar cells are recognized, even though the cells look visually unchanged.

All that's left to do is remove the red grid that helped us understand the problem.

image


inson1:
@KBenMessaoud SMT apps were sold. Fork is here https://github.com/FossifyOrg Please create PR there

Issue #2248, opened 2023-10-31
Option to remove old events


szpak:
The `Error: java.lang.IllegalStateException: Maximum limit of concurrent alarms 5...` (#1805/#1644) problem is very annoying for me. I wonder if removing some old event would help?

If yes, there could be an option to remove the events older than X. It could be done by exporting events, their edition in an external tool and subsequent import, but there might be some metadata missing in that operation. Therefore, it could built-in into the app as one time action called from settings or - as a bonus - a recurring action executed occasionally in the background.

PR #2281, opened 2023-12-05
added option to remove old event in the settings (issue #2248)

MariemMARZOUGUI:
In the settings, it is now possible to remove all tasks and events after a specific date by selecting it.


inson1:
@MariemMARZOUGUI thank, but SMT are dead. There is fork already https://github.com/FossifyX.

PR #2286, opened 2023-12-17
chore(ics): 2024 china holidays

curbengh:
- remove old holidays
- set some holidays as recurring
- add half-day holidays 
- https://www.gov.cn/banshi/2012-11/30/content_2279246.htm 
- https://www.gov.cn/zhengce/content/202310/content_6911527.htm


inson1:
SMT apps were sold. Fork is here https://github.com/FossifyOrg First release in 1-2 weeks

PR #2297, opened 2024-01-16
[L10n] Update MainActivity.kt: "ราชอาณาจักรไทย" -> "ไทย"

bact:
Use a normal form of the country name, not the official. Same as in latest CLDR. http://unicode.org/Public/cldr/44/


inson1:
@bact SMT apps were sold to ad company. Fork is here https://github.com/FossifyOrg
Please open your PR there

bact:
Thank you for the information :)

PR #2349, opened 2025-07-17
fix- due to the gradle and AGP compatibility , we need change gradle …

mohammadrezaberenji:
…version and upgrade it to build project on new Android Studio


JimmysMaths:
@mohammadrezaberenji SimpleMobileTools apps were sold to ad and spyware company. Fork is here:

https://github.com/FossifyOrg

